PR #37, opened 2015-01-21
began adding simple asserts to functions in AutoQC.py

BillMills:
Started adding some simple assertion checks to the functions used in `AutoQC.py`, when I came across something I don't understand. The `assert` I have written for `refereceResults` insists that all reference results be numpy booleans; this assertion raises exceptions, since several of them are of type `<class 'numpy.ma.core.MaskedConstant'>`. This may in fact be correct, but my misunderstanding - @s-good, do you know why we'd be getting masked constants and not `True` or `False` for some of the reference results?

If indeed all is well, let me know before merging this pull, and I'll amend it with a more appropriate assertion test.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -79,9 +79,9 @@ def referenceResults(profiles):
   for profile in profiles:
     refAssessment = profile.t_level_qc(originator=True) >= 3
 
-    #demand reference results returned bools:
+    #demand reference results returned bools, or masked constants for missing values:
     for i in refAssessment:
-      assert isinstance(i, np.bool_), str(i) + ' in reference result list is of type ' + str(type(i))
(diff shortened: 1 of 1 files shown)


s-good:
Hi @BillMills , what appears to be happening is that some profiles contain missing data e.g. there is a depth value but no temperature value associated with it. The missing data are stored in the profile as numpy masked values. In these instances the reference result is also coming back as a masked value rather than True or False. This is working as was intended but if you think this will be awkward it would be OK to change it to return False if there are missing data.

BillMills:
Thanks @s-good, and sorry this has hung so long - currently teaching some workshops in Australia, but I will jump back in here when they're complete.

As for the false vs. masked value, no, you've definitely made the right choice, and it should stay as-is; I just need to write smarter tests. Thanks for clearing that up!

BillMills:
allowed for MaskedConstants per the above discussion.

Issue #28, opened 2014-08-21
Code up EN spike and step check


s-good:
As part of an intercomparison of 'spike tests' (#21) code up the spike and step check described at http://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf pages 20-21. The program file should be called EN_spike_and_step_check.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

Issue #27, opened 2014-08-21
Code up WOD gradient check


s-good:
As part of an intercomparison of 'spike tests' (#21) code up the excessive gradient test described at http://data.nodc.noaa.gov/woa/WOD/DOC/wodreadme.pdf on page 47. The program file should be called WOD_gradient_check.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

BillMills:
Similar question to my first post in #25 - this test is applied to many different parameters; does it make sense to test them separately, or combine them somehow (and if so, how? AND or OR?)

Also, looking ahead - how shall we deal with missing values when evaluating these tests? Let the dataset stand since the specific test couldn't identify a problem with it, or fail it for being incomplete?

s-good:
As with #25, lets just work on the temperature tests at the moment and ignore the others. If there are missing data they should assume to stand rather than failing it, I think.

Issue #26, opened 2014-08-21
Code up Argo gradient test


s-good:
As part of an intercomparison of 'spike tests' (#21) code up the gradient test described at http://w3.jcommops.org/FTPRoot/Argo/Doc/argo-quality-control-manual.pdf on page 8. Where this refers to pressure in db, this can be read as depth in m. The program file should be called Argo_gradient_test.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

Issue #25, opened 2014-08-21
Code up Argo spike test


s-good:
As part of an intercomparison of 'spike tests' (#21) code up the spike test described at http://w3.jcommops.org/FTPRoot/Argo/Doc/argo-quality-control-manual.pdf on page 8. Where this refers to pressure in db, this can be read as depth in m. The program file should be called Argo_spike_test.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

BillMills:
There appear to be two spike tests defined in the reference; the same calculation applied to temperature and salinity. Shall we keep these separate (`Argo_spike_test_temp.py` and `Argo_spike_test_sal.py`), or shall we combine the two spike checks at each depth? The reference has a prescription about what they do when one or both measures show a spike, but how that translates here isn't instantly obvious.

s-good:
For simplicity I suggest we focus on the temperature tests at the moment and ignore salinity so the file can just be named Argo_spike_test.py. We may want to revisit this in the future but I'm not sure at the moment whether it is best to have them separate or in the same file, so I think it best to leave that decision to later.

s-good:
#40 has implemented this test.

PR #38, opened 2015-03-04
Thinking about combining tests

BillMills:
Now that we are effectively examining single tests, one of the big problems left to solve is how to study the *combination* of groups of tests by ANDing and ORing their results. In this PR, I draft some functions to help us explore every possible AND/OR combination of a collection of tests.

`combineTests` in `combineTests.py` is the top-level function for doing this. It takes as an argument a list of lists, where the first index counts tests and the second counts profiles, since this is exactly what we are currently generating around line 206 in `AutoQC.py`. It uses the other functions in `combineTests.py` to enumerate every possible way to combine every possible subset of tests, and returns a list of lists that describe each combination and the resulting assessment of each profile.

There is a dummy execution of this function in `combineTests.py`; running that file will hopefully illustrate what I'm doing.

All these new functions need to be tested before this PR is complete and ready for merge.

**How to read test partitioning syntax**

Tests are collected in groups to AND together, and groups to OR together, and these groupings are expressed in lists. For example, a partition expressed as `[[0,3], [1], [7]]` means (0 AND 3) OR (1) OR (7). Note not all tests need be used in a given partition.


BillMills:
Wrote some tests - but it occurs to me that not all permutations are currently getting generated.

BillMills:
It also occurs to me that there's no current way to express: A and (B or C). But, I think this can be easily fixed by taking two passes at interpreting the structures we're creating: once the way described above, where

[A,[B,C]] = A or (B and C)

and then a second pass swapping the logic:

[A,[B,C]] = A and (B or C)

BillMills:
Addressed permutations issue in 5db74e2, but need to tidy up & write more tests, also think about and/or interpretation as commented above; sometimes necessary, but not always - consider [[A,B]] vs [[A],[B]]; swapping ands and ors would double count in these symmetric cases.

BillMills:
Alright - that ought to do it for a first pass. ToDo:

 - write some more tests (such as #25 #26 #27 #28) so we can see this stuff in action
 - re-think log reporting to accommodate test combinatorics
 - consider how to stop computing time from blowing up rapidly as the number of tests grow.

Issue #22, opened 2014-08-06
Result processing


BillMills:
Now that useful pass / fail statistics are being generated by the testing pipeline, we can start building summary statistics.  In our previous conversations, @s-good described a summary table that compares test results to an expectation for each dataset, one row in which could be generated by the following:

Given the vector of pass / fail information generated for a test by the testing suite, and a list of the same length and format containing the expected pass / fail performance, write a function that returns a 4-element list that counts, in order,
 - number of datasets that passed when they were expected to pass
 - number of datasets that failed when they were expected to fail
 - number of datasets that passed when they were expected to fail
 - number of datasets that failed when they were expected to pass

BillMills:
does #39 sufficiently address this? Close if so.

s-good:
Yes, #39 sufficiently addresses this so this issue can be closed.

PR #39, opened 2015-04-16
Increased number of quality control checks and analysis of results

s-good:
Added two basic quality control checks to allow the combining of tests to be observed. Added new benchmarking code to analyse the results and print out and plot them. This is similar to the functionality described in #22.

NOTE - The two range checks have been changed so that they reject some profiles and are not currently working correctly.

To do - Add the output to the log files rather than print to screen; determine how to cope when the number of tests increase and the number of combinations massively increases.


util/benchmarks.py
```diff
@@ -0,0 +1,91 @@
+import combineTests as combinatorics
+import matplotlib.pyplot as plt
+import numpy as np
+import sag_utilities as sag
```
BillMills: This import doesn't appear to be used (and pip doesn't know what `sag_utilities` is).

BillMills:
Asides from the possibly spurious `sag_utilities` pointed out above, this looks good by eye. Couple things to add to the todo list:

 - Docs to help interpret the figures.
 - Testing in `util/`

But, both those things can get added after merging this PR. Can we safely take out `sag_utilities`?

s-good:
Sorry, the sag_utilities is a local module that I was using for debugging. It can safely be removed.

I've added an extra commit where the import is removed. I've also tidied up the output from the plot_roc function as even with four quality control checks there are too many combinations to list them all. I've added some brief extra comments explaining the plot.

Issue #41, opened 2015-04-20
Fake profile generator


BillMills:
It'd be nice to have a convenient way to generate fake profiles, engineered to intentionally pass or fail different qc-tests; this would help us sanity-check new implementations of qc-tests.

PR #44, opened 2015-04-24
Gradient Tests

BillMills:
This PR contains first passes at the checks in #26 and #27. Notes:

 - Both follow the previous implementation of `Argo_spike_test.py` and its assumptions.
 - `WOD_gradient_check.py` adds a constraint that demands the depths of neighboring measurements not be identical (in order to avoid dividing by 0 in the gradient). But, the text mentions requiring neighboring measurements be separated by at least 3 m in the case of high-resolution instruments; that would also solve the divide-by-0 problem, but it isn't immediately clear to me if this constraint can be required generally, or, perhaps better, if and how the profile object indicates whether it was a 'high resolution' measurement or not.
 - Turned off the verbose logging temporarily, as the combinatorics grow. 
 - `WOD_gradient_check.py` also includes the check described in the text to attempt to distinguish between 0 as a number and 0 as a missing value indicator. Are we accommodating 0 as a missing value indicator? If not, might as well drop this part of the test.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -228,7 +228,7 @@ print('Number of profiles that should have been failed was %i' % np.sum(trueResu
 bmResults = benchmarks.compare_to_truth(combos, trueResult)
 
 # Plot the results.
-benchmarks.plot_roc(bmResults)
+#benchmarks.plot_roc(bmResults)
 
 #logfile
-generateLogfile(verbose, trueVerbose, profiles, testNames)
+#generateLogfile(verbose, trueVerbose, profiles, testNames)
--- a/qctests/Argo_gradient_test.py
+++ b/qctests/Argo_gradient_test.py
@@ -1,0 +1,39 @@
+"""
+Implements the gradient test on page 8 of http://w3.jcommops.org/FTPRoot/Argo/Doc/argo-quality-control-manual.pdf
+"""
+
+import numpy
+
+def test(p):
+    """
+    Runs the quality control check on profile p and returns a numpy array
+    of quality control decisions with False where the data value has
+    passed the check and True where it failed.
+    """
+
+    # Get temperature values from the profile.
+    t = p.t()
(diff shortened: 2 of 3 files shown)


BoyerWOD:
Regarding WOD_gradient_check.py:
> you are correct, separation by 3m should preclude dividing by 0, but the way the 3m minimum works, if depths are less than 3m apart, a value of 3m is used to divide to get gradient/m.  If depths are exactly the same, this is often an indication of problem in the profile, already flagged in the depth inversion check, and the gradient check is skipped - in other words the flagging is dealt with in the depth variable not each individual oceanographic variable associated with that depth.  There is no need to classify an entire profile as high resolution, if two consecutive depths are within 3m, the < 3m criteria should be employed.
> We are not accommodating 0 as a missing value indicator, however, some of our data providers use 0.0 as a missing value without documenting it as such (or simply in error), so it is beneficial to differentiate, if possible, between real measurements of 0.0 and undocumented missing values.

BillMills:
@BoyerWOD thanks! In that case:

 - if two depth measurements are less than 3 m separated, we'll treat them as 3 m apart.
 - the algorithm needs to not fall over when two consecutive depths are exactly the same. In light of what you mention about other tests flagging this scenario, there's no harm in not flagging it here (as is presently done).
 - we'll leave the 0-as-a-missing-data-indicator test in.

I'll implement these changes shortly, unless I hear otherwise.

s-good:
Great! In addition, I noticed a couple of things about the levels that are being flagged:
* In line 36 I think that both qc[i] and qc[i+1] need to be set.
* in lines 40 and 42 I think the index should be i+1 rather than i

BillMills:
that ought to do it - @s-good, I'll leave you to merge this if you think it's a good first pass; now that we have the facilities to build tests to check implementations of qc tests (#41, #45, #46), I'm happy merging this and validating properly in the test suite.

Issue #46, opened 2015-05-06
QC test tests


BillMills:
As discussed in #41 and implemented in #45, we now have a mechanism and first example to create fake profiles, designed to pass or fail a give test, allowing us to build a test suite to ensure the correct performance of the qc tests. Currently, only one trivial example is implemented; behavior of all qc tests should be validated in `qcvalidation.py`.

BillMills:
This work ongoing in #47 - please comment there.

PR #47, opened 2015-05-09
QC test validation

BillMills:
I'll be pushing checks on the qc tests to this branch as I write them over the next little while, as was discussed in #46. I can write checks and profiles that will obviously be flagged by the qc tests to validate gross behavior, but one thing that would be helpful would be if @s-good would describe some more subtle or realistic scenarios of profiles that should be getting flagged. Comment here and I'll implement.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -228,7 +228,7 @@ print('Number of profiles that should have been failed was %i' % np.sum(trueResu
 bmResults = benchmarks.compare_to_truth(combos, trueResult)
 
 # Plot the results.
-benchmarks.plot_roc(bmResults)
+#benchmarks.plot_roc(bmResults)
 
 #logfile
-generateLogfile(verbose, trueVerbose, profiles, testNames)
+#generateLogfile(verbose, trueVerbose, profiles, testNames)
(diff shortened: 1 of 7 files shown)


BillMills:
(the 5 commits above should be merged from #44 - there's enough going on now that we need to start using feature branches).

BillMills:
(notice the `argo_global_range_check` temperature check is failing due to the incorrect lower temperature limit in the test right now, inserted to induce some flags previously).

BillMills:
note c10ae47 nails down a potential floating point error in the implementation of `argo_gradient_test`.

BillMills:
(again, `EN_range_check` tests currently failing due to artificial limit place in the qc test).

BillMills:
In 29bd978, I wrote a validation check for `WOD_gradient_check`'s zero-as-a-missing value indicator, but the way the qc test is currently written, this indicator is redundant; anything that could get flagged by the missing value indicator would have already been flagged by the excessive gradient indicator; makes me think there might be a misinterpretation of the literature in the implementation in `WOD_gradient_check.py` - will examine further.

BillMills:
344aa6b:
 - the tests for `WOD_gradient_check` were mislabeled (swapped terms 'gradient' and 'inversion'); fixed.
 - re: my comments above on 29bd978, on closer inspection the [original WOD document](http://data.nodc.noaa.gov/woa/WOD/DOC/wodreadme.pdf) uses more flags than just our binary pass/fail; it's true that in our case the zero-as-a-missing-value indicator is redundant, so I dropped the validation test for that behavior.

Issue #53, opened 2015-05-15
Code up Argo regional range test


s-good:
Code up the regional range test described at http://w3.jcommops.org/FTPRoot/Argo/Doc/argo-quality-control-manual.pdf on page 6. The program file should be called Argo_regional_range_test.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

BillMills:
#55 has a go at this - note though, that this introduces several dependencies (pyproj, shapely, geos) for the sake of deciding whether a given lat/long is interior to the polygons defined in the text.

Issue #52, opened 2015-05-15
Code up Argo impossible location test


s-good:
Code up the impossible location test described at http://w3.jcommops.org/FTPRoot/Argo/Doc/argo-quality-control-manual.pdf on page 6. The program file should be called Argo_impossible_location_test.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

Issue #49, opened 2015-05-15
Code up EN constant value check


s-good:
Code up the constant value check described at http://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf page 7. The program file should be called EN_contant_value_check.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

PR #55, opened 2015-05-17
New QC tests

BillMills:
will be pushing attempts at the more straightforward of the new QC tests to this PR (more sophisticated ones will get their own feature branch).

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -3,6 +3,7 @@ import json, glob, time
 import matplotlib.pyplot as plt
 import numpy as np
 import os
+import sys
 import util.combineTests as combinatorics
 import util.benchmarks as benchmarks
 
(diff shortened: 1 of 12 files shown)


BillMills:
3f2b658:

Implements the EN_constant_value test from #49; testing validates that this implementation:
 - flags every element in a profile where 9/10 temperature measurements are the same
 - does not flag anything in a profile where 8/10 measurements are the same
 - does not flag anything in a profile where 9/10 measurements are the same, but those measurements are spread over less than 100 m.

Note only temperature (and not salinity) is examined here, per the other qc tests so far.

s-good:
Hi @BillMills - looks good! I'm not sure what would happen if there were missing data values, though. It may just work but I think it would be good to include the check that there are valid temperature and depths at the levels as done e.g. in Argo gradient check just in case it doesn't.

BillMills:
@s-good - agreed, implemented in b7be6f6; also, generalized testing profiles to support missing values; should include missing value validation for all qc tests. More to come shortly.

BillMills:
Argo_impossible_date doesn't have it's time check implemented yet, need to resolve #62 first.

BillMills:
This PR implements solutions for #49, #51, #52 and #53, including validation checks for all of the above. The only outstanding task is resolving #62, so we can finish implementing the time check in the impossible date test (#51) - otherwise this is mergable pending feedback.

Issue #56, opened 2015-05-21
Code up the EN background check


s-good:
Code up the background check on reported levels described at http://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf page 9 onwards. The program file should be called EN_background_check.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

As described in the docs for the EN_spike_and_step_check #28, the results from part of that code should be used by this check, so the relevant code needs to be extracted and put into the background check.

Note that this check requires a lot of auxiliary information in the form of climatology information and error statistics.

BillMills:
I'd like to keep EN_spike_and_step and EN_background implemented independently; in order to assess the reprieves called for in spike_and_step, might it be possible to call EN_background internally there, and use the results to reprieve suspect flags in spike_and_step? That way, we're just calling one function from another rather than pulling pieces apart - unless there's something more subtle I'm not appreciating.

s-good:
Yes, sounds good. I think it might work best the opposite way, though, so that the spike and step check is called from the background check. Could add an argument to tell it to return the suspect levels instead of the rejected levels.

BillMills:
Test implementation looks good, thanks @s-good! Once we test this, we can close out here.

BillMills:
Hi @s-good - I sat down today to write some tests to wrap this issue up, but now I notice we need a `data/EN_bgcheck_info.nc` - can you please add that to the repo (assuming it's a static data file that makes sense to add ...?)

s-good:
Hi @BillMills, the file can be downloaded from http://www.metoffice.gov.uk/hadobs/en4/data/EN_bgcheck_info.nc. The data are free for use for private study and research (http://www.metoffice.gov.uk/hadobs/en4/terms_and_conditions.html).

BillMills:
#92 finishes refactoring `EN_background` and testing all of its helper functions; however, this test is relatively complicated to construct profiles for that will be flagged by it in a full integration test. @s-good, can you independently identify a profile or two that should be flagged by this test?

BillMills:
#96 implements some additional tests for this; after merging that and #92, I think we're in good shape to close out here.

PR #57, opened 2015-05-21
EN spike and step check

s-good:
This is a first attempt at #28. More testing is needed and the code will need to be changed in future when #56 gets implemented.

### DIFF
--- a/qctests/EN_spike_and_step_check.py
+++ b/qctests/EN_spike_and_step_check.py
@@ -1,0 +1,129 @@
+""" 
+Implements the spike and step check used in the EN quality control 
+system. 
+
+The EN quality control system is actually more complicated than is represented
+here. Rather than directly reject levels that are marked as steps, it 
+marks them as suspect and then they are subjected to an extra test (a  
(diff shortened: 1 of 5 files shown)


BillMills:
Looks good Simon - now that qc test validation is up and running, we're safe to merge new qc tests early and try to poke holes in them as we build up the validation tests - I'm going to merge this now and devise some tests shortly thereafter.

qctests/EN_spike_and_step_check.py
```diff
@@ -0,0 +1,129 @@
...
+            wt1 = 0.5
+        
+        # Define tolerance for use later.
+        if (np.abs(p.latitude()) < 20.0): 
+            depthTol = 300.0
+        else:
+            depthTol = 200.0
+        if z[i-1] > 600.0:
+            tTolFactor = 0.3
+        elif z[-1] > 500.0:
```
BillMills: should this be `z[i-1]`?

s-good: Yes, it should be z[i-1]. Thanks for spotting that.

PR #59, opened 2015-05-22
Updates to allow running the tests on a very large number of profiles…

s-good:
… without running out of memory. Also a couple of changes to make the EN_spike_and_step_check more robust and to improve the output from the benchmarking code.

Results from the 150 profile test dataset available in the repository are the same as before. Results when running on the full test dataset are:

155794 profiles will be read
7 quality control checks will be applied:
 Argo_global_range_check
 Argo_gradient_test
 Argo_pressure_increasing_test
 Argo_spike_test
 EN_range_check
 EN_spike_and_step_check
 WOD_gradient_check
Number of profiles tested was 155787
Number of profiles that failed Argo_global_range_check was 2560
Number of profiles that failed Argo_gradient_test was 4331
Number of profiles that failed Argo_pressure_increasing_test was 7514
Number of profiles that failed Argo_spike_test was 1031
Number of profiles that failed EN_range_check was 2351
Number of profiles that failed EN_spike_and_step_check was 7198
Number of profiles that failed WOD_gradient_check was 8201
Number of profiles that should have been failed was 18174

Previously the code would not run on this number of profiles without running out of memory.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -3,6 +3,7 @@ import json, glob, time
 import matplotlib.pyplot as plt
 import numpy as np
 import os
+import sys
 import util.combineTests as combinatorics
 import util.benchmarks as benchmarks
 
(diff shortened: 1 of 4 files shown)


BillMills:
I'd like to find a way to make the full dataset available for reproduction of these results; there are a number of technical options, but are there any copyright or privacy protections preventing distribution of the dataset?

PR #61, opened 2015-05-23
Testing en spike and step

BillMills:
Finishes the first pass at testing `EN_spike_and_step` I started in #58. Notice the build is currently breaking in two places on condition C (steps):

 - a step is getting flagged when it should be permitted by the interpolation condition C(i)
 - a step at the end of the profile isn't getting flagged at all, breaking the test for condition C(iii). I *think* this is because the indices in `conditionC()` are one behind where they should be - meaning they check everything correctly, except one step behind in the loop, meaning the index never actually hits the last element in `dt`.

Will push another commit to this branch with my fixes for these, but wanted to keep it separate unless @s-good can let me know why things are actually correct the way they are.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -3,6 +3,7 @@ import json, glob, time
 import matplotlib.pyplot as plt
 import numpy as np
 import os
+import sys
 import util.combineTests as combinatorics
 import util.benchmarks as benchmarks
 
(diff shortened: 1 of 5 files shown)


BillMills:
c06e579:

Moves the indices as I suggested and checks for being at the end of the profile - Ciii test passes.

BillMills:
aca14da:

adds the interpolation check to avoid rejecting steps, per rule C(i).

(one thing to note - the test for this also illustrates that a smooth decline to the end of the profile, if steep enough, will look like a 'step' - not a problem I guess, since it's consistent with the text - just a bit of a weird edge case).

BillMills:
oh wait, I see what you're doing - that offset index I moved in c06e579 that misses steps at the end is supposed to be handled by that block 

```
    # Step or 0.0 at the bottom of a profile.
    if isData[-1] and dt.mask[-1] == False:
        dTTol = determineDepthTolerance(z[-1], np.abs(p.latitude()))
        if np.abs(dt[-1]) > dTTol:
            qc[-1] = True
    if isTemperature[-1]:
        if t[-1] == 0.0:
            qc[-1] = True
```

which isn't doing it's job atm. Let's do it your way - I'll put the indices back how you had them (since just changing them in C puts them out of step with everything else anyway....) and get that block quoted above to pass the test.

BillMills:
Alright, that ought to do it - the factorization is a bit odd around Ciii, but the build is passing and this branch is ready for merge.

I'd still like to understand the `wt1` parameter a bit more; it doesn't seem to be part of EN_spike_and_step as written, unless I'm just missing it. I could probably refactor the whole function so we don't need an edge case for trailing steps, but I'm reluctant to do so without understanding this additional calculation; but now I'm picking on *really* small things.

Issue #62, opened 2015-05-23
wod decoding: missing data edge case


BillMills:
When some data is missing from a profile's primary header, the corresponding key is never inserted into the primary header dictionary by `_interpret_data`.

As a result, functions like
```
    def time(self):
        """ Returns the time. """
        return self.primary_header['Time']
```

return a `KeyError` when they go looking for them; reproduce this by inserting a call to `p.time()` in any of the qc tests and running on the testing data.

We could solve this by putting exception handling into each downstream function, but I would recommend keeping the keys for every primary header dictionary always exactly the same, and assigning a value of `None` when something is missing. That way everything remains as consistent as possible - thoughts, @s-good?

s-good:
@BillMills - please could you check if I've solved the problem satisfactorily?

BillMills:
the new `wod_tests.py` checks behavior for calling `p.time()` and nothing goes too wrong, so I'm going to call this fixed. Thanks @s-good !

PR #63, opened 2015-06-03
EN background check

s-good:
Implements the EN background check #56. NB. it requires a data file to operate. I need to work out how to distributed the data file. If it is not present the test is effectively ignored.

This also includes an attempt to solve #62.

### DIFF
--- a/qctests/EN_spike_and_step_check.py
+++ b/qctests/EN_spike_and_step_check.py
@@ -2,20 +2,22 @@
 Implements the spike and step check used in the EN quality control 
 system, pages 20-21 of http://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf
 
-The EN quality control system is actually more complicated than is represented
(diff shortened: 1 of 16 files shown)


s-good:
I am merging this straight away as it includes code changes that will be useful in the coding sprint. More work is required to add code tests and to make the auxiliary data file available.

qctests/EN_spike_and_step_check.py
```diff
@@ -62,12 +64,12 @@ def test(p, *args, **kwargs):
         if (np.abs(p.latitude()) < 20.0 and z[i-1] < 1000.0 and
             t[i-1] < 1.0):
                dt[i] = np.ma.masked 
-               qc[i-1] = True
+               if suspect == True: qc[i-1] = True
```
s-good: In the code this is treated as a suspect observation. I guess in the text preliminary rejection must mean suspect.

PR #77, opened 2015-06-05
Impossible date time

BillMills:
Implemented the time check on `Argo_impossible_date` from #51 - but, I'm a little unclear on how the wod format is encoding times. It looks like times are floats on the range 0 to 24, so the condition in the [text](http://w3.jcommops.org/FTPRoot/Argo/Doc/argo-quality-control-manual.pdf) that requires a plausible hour and minute are interpreted here as the time returned from the profile being on that range. Let me know if there's a problem with this interpretation.

### DIFF
--- a/qctests/Argo_impossible_date_test.py
+++ b/qctests/Argo_impossible_date_test.py
@@ -29,5 +29,7 @@ def test(p, **kwargs):
         qc[0] = True
     elif day not in range(1, calendar.monthrange(year, month)[1] + 1):
         qc[0] = True
+    elif time is not None and (time < 0 or time >= 24):
+        qc[0] = True
 
     return qc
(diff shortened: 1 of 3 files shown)


s-good:
Do you see times that are exactly 24? I would have thought that this should not be valid as it is equivalent to 0 hours the next day.

BillMills:
Right you are @s-good - bda015b restricts the time *t* unpacked from the profile to be 0 <= *t* < 24.

BillMills:
Just checking in on this - this is ready for merge if it seems reasonable to @s-good; there's no hurry if you just haven't had time to have a look yet, but if something is unsatisfactory, let me know and I'll fix it right away!

Same comment for #92.

BillMills:
this has hung for a couple of months now - I assume there are no problems, so I'm going to merge, feel free to reopen and comment if something needs attention.

Issue #60, opened 2015-05-22
Combinatorics intelligence


BillMills:
The brute-force combinatorics examinations implemented in #38 are acceptable for small numbers of tests, but compute time will diverge very badly as the number of tests grows. We need a stronger strategy.

The numbers reported in #59 make the individual tests seem much too permissive on their own; one idea could be to look for tests that flag disjoint sets of profiles, and OR them all together.

BillMills:
After looking into machine learning strategies in greater depth, this [paper by Sadohara](http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=ED6330AC1297BAAA56EF7FB507AFDFF3?doi=10.1.1.103.7219&rep=rep1&type=pdf) suggests that a support vector machine (SVM) might be the best choice for this problem. Sadohara's paper examines learning boolean functions (which is exactly what we have - whether a profile passed or failed each individual test are our boolean inputs, and whether the profile should ultimately pass or fail is the boolean output) by examining all possible conjunctions of boolean inputs (which is exactly what we were naively doing in #38, albeit in a much cruder fashion). Sadohara presents two SVM implementations that:

 - produce a lower misclassification rate than other leading techniques for the same amount of training data.
 - remain computationally efficient even when considering a large number of inputs (QC tests).

Furthermore, scikit-learn supports [SVM with custom kernels](http://scikit-learn.org/stable/modules/svm.html#custom-kernels), meaning this ought to be not very difficult to implement.

BillMills:
cc @s-good @BecCowley @BoyerWOD 

#101 presents a [detailed report](https://github.com/IQuOD/AutoQC/blob/master/learningAnalysis/basic-machine-learning/Untitled.ipynb) on a first pass at using machine learning techniques to perform final classification. Executive summary:

 - the most effective techniques correctly flag about 55% of profiles in the full quota dataset that ought to be flagged.
 - corresponding false positive rates on datasets that should not be flagged are around 10%
 - execution time for the full quota dataset on a low-end machine are under a minute. 

These results are preliminary and much more remains to be done to optimize them; comments and suggestions very welcome.

BillMills:
One thing that comes to mind that will produce systematic biases in any final decision strategy is the nature of flagged profiles in the dataset trained on; for example, the machine learning techniques explored above offer only a small improvement over just using `EN_background`; if the quota sample overrepresents profiles flagged by one test (like `EN_background`), this will bias any strategy.

[...]

Issue #48, opened 2015-05-10
Identify a body of data to analyze


BillMills:
Since a usable version 1 of this project is on the horizon, it would be good to identify a body of data we'd like to analyze in early studies.

s-good:
Hi @BillMills , we have a large test dataset that we can use (>500 MB of data) which the 150 profile test dataset was extracted from. I've had to make a few modifications to the code to allow it to run without running out memory, which is not quite ready yet.

BillMills:
Now that #59 is merged (thanks @s-good!), as I commented in that PR it would be good to make this data available if there aren't any problems with distribution rights.

PR #79, opened 2015-06-07
first pass at combining test results with a decision tree

BillMills:
This PR uses scikit-learn to implement a very simple decision tree to assemble individual test results into a final pass or fail for each dataset; thanks to @daisieh at the Mozilla Global Sprint for suggesting this.

Machine learning solutions will resolve the performance problems mentioned in #60 very effectively, but it's difficult to assess them at the moment, since I don't have the full dataset discussed in #48. If that data can be made available, I can do more to refine this approach.


BillMills:
Given what we've learned about machine learning techniques for this problem (discussed in #60), I'm going to close this PR and send a new one using support vector machines instead of decision trees.

Issue #50, opened 2015-05-15
Implement WOD range check


s-good:
Code up the range test described at http://data.nodc.noaa.gov/woa/WOD/DOC/wodreadme.pdf on page 47. The program file should be called WOD_range_check.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

This requires reading in a data file that is used each time the quality control check is run. It would therefore be most efficient to read this in once in the main program and pass the data to the quality control check.

BillMills:
Resources for composing this test:
 - [lat / long associated with region code](http://data.nodc.noaa.gov/woa/WOD/MASKS/range_area.msk)
 - [region code associated with region name](http://data.nodc.noaa.gov/woa/WOD/MASKS/range_basin_list.msk)
 - [region name associated with max / min temperatures by depth](http://data.nodc.noaa.gov/woa/WOD/DOC/wodreadme.pdf), appendix 9.1

Does appendix 9.1 listed above exist as a CSV somewhere?

s-good:
Hi @BoyerWOD - is appendix 9.1 from the WOD readme doc available as a CSV file?

BillMills:
@BoyerWOD yes please! If an official csv or json of that table is available, that would be ideal; #78 extracts it well enough for development to proceed, but the validation there is not up to production code atm.

BoyerWOD:
Please find the following min/max range files:
ftp://ftp.nodc.noaa.gov/pub/WOD/SELECT/WOD_ranges_Temperature.csv
ftp://ftp.nodc.noaa.gov/pub/WOD/SELECT/WOD_ranges_Salinity.csv
ftp://ftp.nodc.noaa.gov/pub/WOD/SELECT/WOD_ranges_Oxygen.csv
ftp://ftp.nodc.noaa.gov/pub/WOD/SELECT/WOD_ranges_Nitrate.csv
ftp://ftp.nodc.noaa.gov/pub/WOD/SELECT/WOD_ranges_Phosphate.csv
ftp://ftp.nodc.noaa.gov/pub/WOD/SELECT/WOD_ranges_Silicate.csv

Note that ranges have been updated in some cases from the User Manual
and new geographic areas have been added in the Northwest Pacific in conjunction with our regional atlas of the East Asian Seas.

BoyerWOD:
[...]

Issue #54, opened 2015-05-15
Code up EN track check


s-good:
Code up the constant value check described at http://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf page 7. The program file should be called EN_track_check.py and be put into the qctests directory. It should follow the structure shown in EN_range_check.py.

This is quite a challenging check to code up as the check needs to use information about other profiles rather than just the profile itself. Therefore the whole list of profiles will need to be passed in as an additional argument.

BillMills:
I'll have to look a bit closer, but instead of passing as an argument, we could load the list of profile headers (just the headers, not the level data) as part of the global datastore being introduced in #84 for `WOD_range`; that way, we avoid having to add special cases or extra kwargs to the main routine to support individual tests. Not necessarily better - just an alternative that I think might help avoid blowing up the complexity of the main routine in future.

BillMills:
resolved in #93.

PR #84, opened 2015-06-15
first pass at WOD_range_check

BillMills:
here's a first pass at #50, including the necessary data tables and helper functions to consume them. Test passes a very simple validation check, but I'd like to add some more tests before signing off on this one.


BillMills:
Alright, added some more tests & that ought to do it - we can close out #50 once this is merged.

BillMills:
I'm going to go ahead and merge this in, since I'm thinking about using some of the data store infrastructure for `EN_track_check` in #54.

PR #93, opened 2015-06-20
EN_track_check

BillMills:
I'll be pushing progress on #54 to this pull request; still a long way to go, but some of the strategy is visible right now.

Unlike all the other qctests, `EN_track` requires all the profiles from a given cruise all at once. But, for performance, we don't want to load any given profile more than once. In order to approach this, I'm trying the following:

 - after loading the full list of profile primary headers, sort by cruise number.
 - instead of running the full suite of tests after loading each individual profile, load an entire cruise worth of profiles, make them available on the global data store, then run each test individually exactly as previous, profile by profile.
 - have `EN_track` write a copy of its results to the global data store. This way, the first time `EN_track` is called for a profile in a cruise, it can pull all the relevant profiles from the data store, run the qc check for the whole cruise, and on each subsequent call for the rest of the profiles in the cruise, the answer can be fetched and returned from the data store without actually rerunning the qc test.

The actual guts of `EN_track` aren't implemented yet, but just running this as-is seems to make the correct information available at the correct times, without any additional reads, and without adding any specifics of `EN_track` to the main routine. 

Some concerns:

 - sorting the profile primary headers by cruise number could get slow for a huge number of profiles. Can we perhaps safely assume that all the profiles for a given cruise are in the same file? Sorting a bunch of small lists is much faster than sorting one huge list.
 - This whole scheme assumes it won't be too cumbersome to load all the profiles for a single cruise. Good assumption / bad assumption?

@s-good: I'm going to forge ahead with this strategy if I don't hear from you; let me know if you anticipate any pitfalls here.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -33,9 +33,11 @@ def processFile(fName):
   firstProfile = True
   currentFile  = ''
   f = None
-  for iprofile, pinfo in enumerate(profiles):
-    # Do not process data twice if run in parallel.
-    if pinfo.file_name != fName: continue 
+
+  # keep a list of only the profiles in this thread
+  data.ds.threadProfiles = main.extractProfiles([fName])
+
+  for iprofile, pinfo in enumerate(data.ds.threadProfiles):
     # Load the profile data.
     p, currentFile, f = main.profileData(pinfo, currentFile, f)
     # Check that there are temperature data in the profile, otherwise skip.
--- a/qctests/EN_track_check.py
+++ b/qctests/EN_track_check.py
@@ -15,8 +15,8 @@ import math
 DistRes = 20000. # meters
 TimeRes = 600. # seconds
 
-ds.EN_track_headers = {}
-ds.EN_track_results = {}
+EN_track_headers = {}
+EN_track_results = {}
 
 def test(p):
     """ 
@@ -35,25 +35,25 @@ def test(p):
     # qc return objects) 
 
     # check if this profile has been examined already
-    if (cruise, uid) in ds.EN_track_results.keys():
-        return ds.EN_track_results[(cruise, uid)]
+    if (cruise, uid) in EN_track_results.keys():
+        return EN_track_results[(cruise, uid)]
 
     # some detector types cannot be assessed by this test; do not raise flag.
     if p.probe_type in [None,0]:
         return np.zeros(1, dtype=bool)
 
-    # the first time this test is run, sort ds.profiles into a cruise-keyed dictionary:
-    if ds.EN_track_headers == {}:
-        ds.EN_track_headers = main.sort_headers(ds.profiles)
+    # the first time this test is run, sort ds.threadProfiles into a cruise-keyed dictionary:
+    if EN_track_headers == {}:
+        EN_track_headers = main.sort_headers(ds.threadProfiles)
 
     # since we didn't find an answer already calculated,
     # we still need to do the calculation for this cruise;
-    # all the relevant headers are sitting in the ds.EN_track_headers list.
-    headers = ds.EN_track_headers[cruise]
+    # all the relevant headers are sitting in the EN_track_headers list.
+    headers = EN_track_headers[cruise]
 
     # start all as passing by default:
     for i in range(len(headers)):
-        ds.EN_track_results[(headers[i].cruise(), headers[i].uid())] = np.zeros(1, dtype=bool)
+        EN_track_results[(headers[i].cruise(), headers[i].uid())] = np.zeros(1, dtype=bool)
 
     # copy the list of headers;
     # remove entries as they are flagged.
@@ -67,9 +67,9 @@ def test(p):
     # if more than half got rejected, reject everyone
     if len(passedHeaders) < len(headers) / 2:
         for i in range(len(headers)):
-            ds.EN_track_results[(headers[i].cruise(), headers[i].uid())][0] = True        
+            EN_track_results[(headers[i].cruise(), headers[i].uid())][0] = True        
 
-    return ds.EN_track_results[(cruise, uid)]
+    return EN_track_results[(cruise, uid)]
 
 def findOutlier(headers):
     '''
@@ -95,7 +95,7 @@ def findOutlier(headers):
     if flag:
         rejects = chooseReject(headers, speeds, angles, iMax, maxSpeed)
         for reject in rejects:
-            ds.EN_track_results[(headers[reject].cruise(), headers[reject].uid())][0] = True
+            EN_track_results[(headers[reject].cruise(), headers[reject].uid())][0] = True
         return rejects
     else:
         return []
@@ -134,7 +134,8 @@ def detectExcessiveSpeed(speeds, angles, index, maxSpeed):
     '''
 
     flag = speeds[index] > maxSpeed
-    if index < len(angles) - 1:
+
+    if index < len(angles) and index > 0:
         flag = flag or ( (speeds[index] > 0.8*maxSpeed) and (angles[index]>math.pi/2 or angles[index-1]>math.pi/2) )
 
     return flag
--- a/qctests/EN_track_check.py
+++ b/qctests/EN_track_check.py
@@ -205,7 +205,7 @@ def condition_a(headers, speeds, angles, index, maxSpeed):
             return 0
     elif index == len(headers)-1 and len(headers)>3:  # why not >=? seems to cause problems, investigate.
         impliedSpeed = trackSpeed(headers[-3], headers[-1])
-        if impliedSpeed < maxSpeed and (speeds[-2] > maxSpeed or angles[-2]>45./180.*math.pi):
+        if impliedSpeed < maxSpeed and (speeds[-2] > maxSpeed or angles[-3]>45./180.*math.pi):
             return index-1
         else:
             return index
--- a/qctests/EN_track_check.py
+++ b/qctests/EN_track_check.py
@@ -173,6 +173,32 @@ def detectExcessiveSpeed(speeds, angles, index, maxSpeed):
 
     return flag
 
+def meanSpeed(speeds, headers, maxSpeed):
+    '''
+    determine mean speed, neglecting missing data, intervals less than 1h, and speeds over maxspeed, for use in condition (f)
+    '''
+
+    meanSpeed = 0
+    speedCount = 0
+    for iSpeed, speed in enumerate(speeds):
+        if speed == None or iSpeed == 0:
+            #missing values
+            continue
+        elif iSpeed > 0 and geo.deltaTime(headers[iSpeed-1], headers[iSpeed]) < 3600.:
+            #too close together in time
+            continue
+        elif speed > maxSpeed:
+            #too fast
+            continue
+        else:
+            meanSpeed += speed
+            speedCount += 1
+
+    if speedCount > 0:
+        meanSpeed = meanSpeed / speedCount
+
+    return meanSpeed
+
 
 def trackSpeed(prevHeader, header):
     '''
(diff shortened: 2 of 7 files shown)


s-good:
Hi @BillMills just to clarify, we don't need to load the full profile data for this test. The information that is loaded with the load_profile_data keyword set to False in the WOD profile reader is adequate, so there shouldn't be any memory issues.

BillMills:
@s-good That's much simpler then! I'm at a conference all week, but I will change course hopefully this weekend. Thanks for having a look!

BillMills:
Today's commits do a few things:

 - sends the headers (and only the headers) to the data store, sorted by cruise and time
 - implements a basic speed check between time-adjacent points on a cruise.
 - note that the text wasn't very explicit about their `Dist(K)` variable; this seems to be the distance between measurements, which I take as the haversine of the longitudes and latitudes of adjacent measurements.

This has the data routing and sorting nailed down, but the test itself is currently a toy implementation only. Still to do:

 - Distinguish between ship and buoy measurements.
 - Implement angle calculations between track segments
 - Implement logic to decide which endpoint of a track segment is incorrect, when one is flagged
 - Tests.

BillMills:
aad1bf3 approaches a complete implementation of EN_track, except for a few things I need clarification on from @s-good :

 - Condition (a) in the text mentions that 'An analogous test is used if M is the last report in the track', but the analogy is not obvious, since condition (a)'s test refers to the speed at position M+1.
 - Condition (e) appears to have a misprint in the text; there is no boolean operator anywhere in the expression as written.
 - Condition (h) has the same problem as (e).
 - It's not immediately obvious how to decide if a track is a ship or a buoy; I see the `probe_type` function on the WOD class, and table 29 on page 90 of the [WOD docs](http://data.nodc.noaa.gov/woa/WOD/DOC/wodreadme.pdf), but I don't quite have the domain knowledge to parse that as 'ship' or 'buoy'.

Once those are cleared up, we can check the implementation with some tests and merge afterwards.

castelao:
In respect to the last point, from table 29

1, 2, 5, 6 & 8 should be ship
14 could be from ship as well
12 is ship, but reduced speed
4, 7 are ship, but drifting, i.e. not underway
10 is fixed position, moored
9, 11, 13, 15 slow speed

BillMills:
Hi @castelao, thanks so much! And sorry for the absurdly slow reply, I've been traveling for a few weeks.

So, this helps a lot, but there are some issues; for example, does this test even make sense for 10 (fixed position)? Should we opt out for those profiles in that case? For 4, 7 and 12, I'm going to assume the slow speeds mean we should use the buoy characteristic speed, even though these are ship-based measurements. Under these assumptions, we have:

code | type | max speed (m/s)
----|---|-------
1 | MBT | 15
2 | XBT | 15
3 | DBT |
4 | CTD | 2
5 | STD | 15
6 | XCTD | 15
7 | bottle / rossette / net | 2
8 | underway / intake |15
9 | profiling float | 2
10 | moored buoy | (abort test)
11 | drifting buoy | 2
12 | towed CTD | 2
13 | animal mounted | 2
14 | bucket | 15
15 | glider | 2
16 | microBT |

Somehow I suspect some more of these should be dismissed (gliders? animal mounts?) as inappropriate data sets for this test - @castelao @s-good, help me fill the rest of this out and we can knock one of the above bullet points off the list.

castelao:
Hi, don't worry.

Those speeds sound reasonable for me.

I would keep gliders and animal mounts, so it could detect typos or other mistakes in the position and/or timing.

Might be valid the test for 10, maybe not exactly the same test. A mooring can break free and drift for a while before being recovered. Other possibility is to transmit, or store, a wrong GPS position or time. Such test would capture that. But a 0 velocity wouldn't work due the noise in the GPS position.

BillMills:
Okay great - buoy / ship discrimination is implemented in 9f8f85a as described in the table above; note that for DBT and microBT, no comment was made, so these abort the test (ie no flags raised) at the moment. Let me know any changes requested to the above table. Thanks @gui!

s-good:
Hi @BillMills, sorry for the extremely delayed response to this:

First bullet: if the last three points in the track are x, y, z then if speed between x and z < tolerance and (speed between x and y > tolerance and angle at x > 45) then reject y else reject z.

Second bullet: > signs seem to have been missed out - e should read If Angle(M − 2) > 45 and Angle(M − 2) > Angle(M + 1) then reject M − 1; else if Angle(M + 1) > 45 then reject M.

Third bullet: same as above; h should read If ABS(PD1 − PT1) > 0.1 + ABS(PD2 − PT2) then reject M − 1; if ABS(PD2 − PT2) > 0.1 + ABS(PD1 − PT1) then reject M.

DBT and microBT can probably be treated the same as MBT as they are grouped with the MBTs in the WOD.

Many thanks for coding up this complicated check!

BillMills:
Thanks, @s-good - I'll revisit this check soon, but first we're going to have to smarten up our parallelization strategy; currently the script that splits our datasets up just blindly separates them into n chunks of equal size, but we'll need to make sure every profile from a given track is in the same chunk in order for this test to work in our infrastructure.

Looking ahead, can you foresee any other qc tests that will necessitate grouping together particular profiles in a single subset?

s-good:
One of the EN checks is a buddy check which compares a profile to other, nearby profiles. The requirement is different to the track check as it requires profiles recorded using different instruments rather than all from the same platform/instrument for the track check. One option is to split the data into boxes according to profile location, but we'll still get problems at the edges of the boxes and the problem would get worse the more parallel threads we use. A better option would be to split by time (which may be what is already happening as WOD files are normally in date order but I'm not sure if this is the case for QuOTA). Unfortunately, though, it is possible that some QC checks will have a problem with this, too, although I don't know of any examples at the moment so we may be OK.

BillMills:
So, as discussed, I was looking into a new data splitting strategy that preserves cruise groupings for the track check; the latest commit provides a simpleminded strategy that does ok in the limit of total data >> single cruise data - which would have worked great if it wasn't for cruise 900229. This cruise appears to have just over 67000 measurements attributed to it - an order of magnitude more than anything else in *quota*. I guess it's possible but... seems suspicious? 

If it's correct, that's a serious problem for parallelizing the track check, since we can't go any faster than a single core chewing through the largest subset.

castelao:
I don't know about this 900229 cruise, but there are some repeating vessels of opportunity that have been measuring for a while, like the Oleander.

Maybe split the data in blocks with a "buffer" zone from the adjacent group. The data in the buffer is not evaluated, but only for support.

A smart data split would require to know in advance what it has to be evaluated. Another idea would be QC in two fronts. In the first one sequentially distribute the data for the regular tests that are independent, while collecting date/lat/lon/uid. Maybe an sqlite? While the regular tests are being processed, the sqlite is consumed applying the speed/track check, and searching for match ups for the buddy check. In my experience the heavy load is on defining the match ups, which run faster with indices. I did the buddy check before with all the data in a PostgreSQL + PostGIS, in a different scenario, probably not the best here.

s-good:
It's possible e.g. that this a default code for if the cruise number is unknown, in which case the track check shouldn't be applied since there could be multiple ships reported under the same code. The CSIRO people should be able to advise.

BillMills:
@castelao I really like your idea of a two pass method that populates a sqlite database on the first go, so we can fish out groups of profiles (cruises, buddies, whatever) on the second pass efficiently, and allocate them to parallel threads that way. My only concern is that this is a big engineering project and will take (me at least) a long time - @s-good, what's our outlook wrt deadlines?

s-good:
We have until May to implement as many QC checks as possible. I think we are in pretty good shape but we've still got the remaining ICDC and CSIRO tests to implement plus any other tests that we find that we feel should be included (e.g. maybe some more from WOD) so there is still potentially quite a lot to do.

BillMills:
Okay, May is going to come awful fast - I'm going to focus on @s-good's list before I worry about more infrastructural adaptions. But again, the two-pass database is a good idea that should be pursued as time permits.

BillMills:
last push catches us up with the large infrastructural changes since I first started working on this test, but more remains to be done to make sure this is playing nicely with our most current parallelization scheme, and reflects the text correctly.

BillMills:
Parallelization note: the condition introduced in #145 to make sure individual threads don't try to QC the entire list of headers kept on hand for the buddy check was a fine solution previously, but will cause problems with the track check since it uses the header list *inside* the qc test, circumventing the patch in #145; we'll begin keeping an additional header list of just those headers in the current thread, to loop over and for use in the track check.

BillMills:
Alright - I've done all I can think of to validate this test, and it seems to be working; would nevertheless recommend validation against real profiles as per #136 when feasible. Unit testing checks:

 - all helpers
 - each decision step (a)-(h) as evaluated individually (unit tests) and as part of the full chain (integration tests), both for flagging the correct profiles, and in the case of the integration tests, that the flagged profiles are flagged by the expected step in the chain.

This test as implemented trusts that:

 - all profiles on a given track are in the same file / parallel thread
 - the headers of the tracks in the file considered by this thread are found on `ds.threadProfiles`

@s-good, I almost hate to ask you to review such a beast of a pull request, but given the complexity of this test, a second set of eyes comes highly recommended. Perhaps (perhaps) the easiest thing to do is examine the test suite and make sure the things it's asserting are true actually should be true. If you go this route, most of the tests rely on arrays of artificial profiles engineered to have specific problems; they all exploit the same few facts to cause faults in the desired places:

 - 0.5 degrees latitude can be traversed in 1 hour at < 15 m/s, but not 1 degree.
 - almost all tracks are straight lines that may double back on themselves when large angles are needed.

Let me know any problems,

s-good:
OK, will take a look. I will also try to dig out some real world examples to test.

s-good:
I've had a look through the code, which looks fantastic. Thanks so much for doing this - it must have taken ages to do. 

I've added some new tests including real examples which I'll try to send via a pull request. The good news is that the real examples all worked fine. 

I did manage to break the code by having two profiles with identical times. I started trying to figure out how to solve it but realised I would probably end up messing it up by not understanding it well enough. It might not be related, but I noticed in line 172 of the QC check code that there is an if index < len(angles), which I think may always be true? Possibly it should be index < len(angles) - 1? As I say, I didn't want to start introducing changes like this without knowing the code well enough.

s-good:
I also just tried running the 150 test profile set through AutoQC. The track check rejects 83% of good profiles (which is excellent) but also 83% of bad profiles (which is not excellent!). Does this indicate that it is running the track check on a set of profiles that are not from the same instrument and needs to be excluded from the track checking? E.g. it looks like profiles with cruise ID == 0 are being rejected, which should be excluded from track checking I think.

BillMills:
OK, a few comments:

 - The identical times test was erroring previously because we weren't properly handling the case where there were only two profiles on the track (good thing @s-good's new tests thought of that!). Now the test fails instead of errors, because as written it expects the case of two profiles with different positions and identical times to pass, but EN-track sees the second profile as having an infinite velocity, and decides to reject the first profile. Which doesn't sound unreasonable to me; @s-good, what behavior ought we have when identical times are found?

 - I don't think @s-good's point about the index checking around 172 was responsible for the error state discussed above, but I *do* think it was a weird decision in hindsight; if we're at the end of the `angles` array, it'll find `None` and not pass the next conditional, which seems fine, no need to avoid this. On the topic, I also removed the `-1`'s from the front of `speeds` and `angles` and replaced them with the more-appropriate `None`.

BillMills:
Oh and:

 - I'm not taking any of the results on `quota_subset` too seriously; it's just too small a dataset to draw any meaningful conclusions from. I like running on it just to make sure there isn't something spectacularly broken that the unit tests were too atomic to catch, but other than that, we'll see what's what when we run the full stack again (soon - after this test is sorted). @s-good, I agree that running `track` on things with cruise ID == 0 doesn't make sense; if we're concluded there, I'll address that before we merge.

s-good:
Hi @BillMills, it's a good question about what should happen if times are identical. I put in the test more to make sure that the code could handle the situation without breaking and hadn't thought properly about the end result. The code is sort of designed to handle this by having the TimeRes as the minimum time difference that is allowed, so we can probably accept what the code is doing. 

I was thinking about the profiles with probe type 0 (unknown type) and I think that they could maybe be checked assuming the higher of the two max speeds.

Thanks again for doing this!

BillMills:
Alright - implemented everything in this last round of discussions:

 - don't test track id 0
 - assume the higher max speed for probe id 0
 - allow test to make its own decisions about profiles with identical times.

I think this is stable and robust enough to merge - a paltry 10 months after #54 was opened!

PR #96, opened 2015-09-29
Simple combination generator and test for EN background check

s-good:
The partition generator has turned out to be slow when run for large numbers of tests. To enable results to be generated, this change has implemented a simple algorithm to produce a subset of the combinations.

A test for the EN background check has also been implemented. Only a single case is implemented so far; ideally more should be added in the future.

A change to the spike check is also implemented that stops a potential divide by zero error.


s-good:
The automatic testing is failing because the auxiliary data needed by the EN background check aren't kept in the repository. @BillMills - can you advise on how to handle this?

qctests/EN_spike_and_step_check.py
```diff
@@ -167,9 +167,13 @@ def conditionC(dt, dTTol, z, qc, t, i, suspect):
         if z[i-1] <= 250.0 and dt[i-1] < -dTTol and dt[i-1] > -3.0*dTTol:
             # May be sharp thermocline, do not reject.
             pass
-        elif i>1 and np.abs(t[i-1] - interpolate(z[i-1], z[i-2], z[i], t[i-2], t[i])) < 0.5*dTTol:
-            # consistent interpolation, do not reject
-            pass
+        elif i>1 and z[i] - z[i-2] > 0:
```
BillMills: good catch - but this can be combined into one else statement a bit more cleanly. For example:

```
x = 0
if x!=0 and 1/x:
    ....
```
will never evaluate the 1/0, since the first condition fails and preempts execution of the second.

tests/EN_background_check_validation.py
```diff
@@ -0,0 +1,22 @@
+import qctests.EN_background_check
```
BillMills: looks good - btw, there's been a bunch of tests for the background check in #92 for some time now; I'm going to merge them in after this, unless I hear any objections.

BillMills:
Okay, I see the issue - all that's missing is that auxiliary data file. It's about 70 MB, which I would just as soon not add to the repo - but I think I can set Travis to download it before running its tests.

I'm going to merge this as-is, fix the build by getting that download working, and make the trivial corrections I point out in the line comments. Thanks for this, overall it looks really good!

PR #106, opened 2015-10-31
first pass at en-stability

BillMills:
This PR implements & tests EN_stability. A few concerns:

 - [the text](http://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf) describing the test was a bit hard to follow; I assumed that the test for density spikes proceeded if and only if the first threshold failed (delta rho < -0.03 kg/m3) - otherwise, the subsequent spike check would always flag every level.
 - McDougall refers to salinity measured in psu, but the WOD definition document reports salinities as unitless; is there a unit conversion that needs to be done?

In other news - this completes all outstanding qc tests save the track check. The one clear result coming from our early machine learning study, is that we don't have enough qc tests implemented yet to make a final qc decision. With our current set of qc tests, about half of the profiles that should be flagged in *quota* never are. Let me know what other tests should be introduced, so we can get as many of them implemented as possible before Hamburg.

### DIFF
--- a/qctests/EN_stability_check.py
+++ b/qctests/EN_stability_check.py
@@ -15,21 +15,55 @@ def test(p, **kwargs):
     # Get temperature, salinity, pressure values from the profile.
     t = p.t()
     s = p.s()
-    p = p.p()
+    P = p.p()
 
     # initialize qc as a bunch of falses;
-    qc = numpy.zeros(len(t.data), dtype=bool)
-
+    qc = numpy.zeros(p.n_levels(), dtype=bool)
+    
     # check for gaps in data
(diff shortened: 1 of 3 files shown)


s-good:
Hi Bill, thanks for implementing this. Salinity is unitless but PSU is often put as the unit (e.g. https://en.wikipedia.org/wiki/Salinity#Seawater).

s-good:
Hi Bill, just to let you know that there is a converter to potential temperature in util/obs_utils.py. I will merge the pull request for now, though, as the test shouldn't be that sensitive to the way this conversion is done.

PR #111, opened 2015-12-08
added CSIRO gradient test

BillMills:
Implements an interpretation of the CSIRO gradient test described in [the paper](http://journals.ametsoc.org/doi/abs/10.1175/JTECHO539.1) recommended by @BecCowley. A few comments:

 - instead of making any replacements per the text, profiles with suspicious characteristics are simply flagged for further inspection, per our priority of minimizing false negatives.
 - I assumed the wire break criteria only applied to XBT probes, but I may have misinterpreted the text; let me know.
 - the spike check suggested in the text is not yet implemented; the exact algebraic definition of a spike was not immediately obvious from the text. Will try to deduce from the fortran.

This test seems simple enough that attempting to wrap the FORTRAN seemed unnecessary. Interestingly, even though we already have two other gradient checks, this one flags several profiles in the test `quota_subset` the others do not; it remains to analyze a full dataset to determine the nature of this sensitivity.

### DIFF
--- a/qctests/CSIRO_gradient.py
+++ b/qctests/CSIRO_gradient.py
@@ -1,0 +1,55 @@
+"""
+Implements the gradient test of DOI: 10.1175/JTECHO539.1
+All questionable features result in a flag, in order to minimize false negatives
+Spike check is ommitted atm. 
+"""
+
+import numpy
+
+def test(p):
+    """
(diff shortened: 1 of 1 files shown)


s-good:
In future we might want to separate out the shallow XBT and the wire breaks into two further separate tests as then we can see the performance of the gradient check separately but will merge for now.

PR #112, opened 2015-12-08
CoTeDe checks

s-good:
This is a first attempt to implement CoTeDe checks in the AutoQC framework. Any comments on this are appreciated!

I got into a bit of a mess with the main AutoQC.py file as I had put in changes to make the output to screen more useful. This wouldn't automatically merge with the latest version and I couldn't figure out how to get rebasing to work so I've ended up copying in the latest version and then re-adding the changes I made but unfortunately even this won't automatically merge!

I've added in a test for the new CoTeDe QC but this will require the CoTeDe module to run so will currently fail.


BillMills:
Hi @s-good - great, thanks for pulling this together so fast! This will also further collide with infrastructural tweaks in #110, but I will sort out everything and merge both this evening.

BillMills:
Hi @s-good - alright, I've got everything sorted out on my machine, there was nothing too hairy - #110  and #108 affected `AutoQC.py`, but not in any way that would really pathologically affect what you're doing here. 

Anyway, I love the new reporting! The only thing we need to do before we can merge, is get the `cotede_qc` package included in the install scripts so that the tests pass (and so that I can run things myself :) - can you let me know how to get ahold of that package?

I'll send a pull request of my corrections to your `s-good:master`, which should make them appear here (but the tests will still fail until we get that package included per above).

s-good:
Hi @BillMills, the CoTeDe package is available with pip but I found that I had problems with using as it was complaining about a missing module - I haven't had a chance to investigate this further yet, though. Instead I downloaded the code from https://github.com/castelao/CoTeDe and then used the setup.py script to install.

s-good:
Hi again @BillMills , I retried using pip to get CoTeDe and found it worked OK. I added the lines to the travis configuration to install this and the seabird package which is also needed and found that the checks are now passing. Thanks for your help with sorting out the merging of the files!

BillMills:
Fantastic, great work @s-good, looks good to me - and huge thanks to @castelao for pulling together the CoTeDe package, I'm thrilled this is all integrating so smoothly.

castelao:
Soon I'll have some time to help, if you needed it. Let me know if you have any trouble with CoTeDe.

@s-good, that was my bad, I forgot to update pypi with github updates. I'll keep a better track on that. Thanks for integrating CoTeDe into IQuOD.

@BillMills, you're welcome.

PR #118, opened 2015-12-14
Tidying up output when using parallel running

s-good:
The output to screen was not working well when running in parallel as messages were printed to screen for each file that was processed rather than for the full results. These changes are aimed at improving this and the percent process indicator has been removed - I may try to reincorporate this in future as I found it quite useful to track progress.

In the course of doing this I noticed that only the output from the first file processed was being written to the csv file and I've tried to fix this.

I have removed the logfile writing function as this is now unused and was previously found to be very slow.

It may be better in future to read the profile list in the main function like is now done for the list of QC checks so that the QC checks can see all the profiles rather than just the ones in the same file as the profile currently being QCed. This might help avoid the need for careful division of the input files for tests like the track check, although it might complicate things in other ways.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -177,6 +177,11 @@ if len(sys.argv)>2:
   results = parallel_result[0][1]
   tests = parallel_result[0][2]
   profileIDs = parallel_result[0][3]
+  for pr in parallel_result[1:]:
+    truth      += pr[0]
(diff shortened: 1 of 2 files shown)


BillMills:
And that's why we *always* want two sets of eyes on a project - great catch on the csv recombination, I was worried that infrastructure was under-tested... Looks good, thanks for this!

PR #120, opened 2015-12-20
CoTeDe checks

s-good:
This introduces the rest of the CoTeDe checks that we don't already have in AutoQC.


BillMills:
Excellent fix, thanks @s-good, I had been fighting with this all day - one other thing I found while digging around in the guts of cotede, was that things will fail if it moves onto trying to download from the URL provided, and `pydap` isn't installed (then if `pydap` is installed, it'll get mad that we didn't give it a proper URL in the json).

BillMills:
Also more generally, we need to sort out exactly what our dependency stack is - @castelao, you dropped `seabird` since it was a dep of `cotede`, but seabird (and `gsw` and `scikit-fuzzy`) are still in `travis.yml`, and as mentioned above `cotede` is using `pydap` which isn't pulled in automatically anywhere.

castelao:
Since AutoQC uses wodpy, seabird is not necessary.

PyDAP is an optional dependency. It is preferred to access the climatology locally with netCDF4, for speed. One solution is (http://cotede.readthedocs.org/en/latest/install.html)

> from cotede.utils import supportdata
> supportdata.download_supportdata()

and don't bother with PyDAP. In anyways, I'm working to improve the climatologies tests.

gsw and scikit-fuzzy are optional dependencies as well, but they are required to apply density-inversion and fuzzy tests. It's probably a good idea to include them in the requirements.

s-good:
Hi @castelao, thanks for the pointer to the supportdata.download_supportdata() function. This looks a better way to get the data, particularly the etopo5 data. I've been trying it out on my machine but it is hanging. Any thoughts on why this might happen? 

I'll remove the dependence on seabird from the travis build.

BillMills:
@castelao thanks for clearing that up! @s-good yes I agree, if `supportdata` lets us ensure that the etopo file is present so we never fall into `cotede`'s catch that tries to download the data with `pydap`, then we can avoid that extra dependency.

Issue #105, opened 2015-10-29
running AutoQC on AWS


BillMills:
Pushing all of *quota* through AutoQC on my home machine takes ~20 hours; it may behoove us to be prepared to run full data reduction runs on AWS. Here are some notes on how I got AutoQC up and running there:

 - make a [free account](http://aws.amazon.com/) at AWS
 - follow the [instructions to set up access to EC2](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/get-set-up-for-amazon-ec2.html) - I followed these exactly, except for the warnings about not allowing ssh from anywhere; there's no security concerns here, easiest to just allow ssh.
 - After signing in as your IAM user from the last step, go to the 'Services' link in the top navigation bar, and choose:
  - EC2
  - launch instance
  - amazon linux
  - t2.micro
  - launch

After launching this compute cluster, a link is presented to go back to the list of instances; go there, wait for this new instance to finish initializing, click on it, click connect at the top, and follow the instructions to ssh into your new node.

Once there, set up AutoQC and all its dependencies via the following:

```
wget https://3230d63b5fc54e62148e-c95ac804525aac4b6dba79b00b39d1d3.ssl.cf1.rackcdn.com/Anaconda-2.3.0-Linux-x86_64.sh
bash Anaconda-2.3.0-Linux-x86_64.sh
```
(disconnect / reconnect)
```
sudo yum -y install geos
conda install netcdf4
sudo yum -y install git
git clone https://github.com/IQuOD/AutoQC.git
sudo yum -y install gcc-c++ python27-devel atlas-sse3-devel lapack-devel
pip install -r requirements.txt
```

After which AutoQC runs happily on the demo dataset it ships with. Next up will be sticking *quota* on S3, and parallelizing. Should be simple enough to slice *quota* up into pieces and run in parallel - nothing too fancy should be required.

Note this service is not free - the setup demo I described above should (should) run on free-tier resources, but a large cluster to run on all of *quota* will cost a few dollars; hopefully this will be in the single digits and will only have to be run very rarely to reduce the full raw dataset.

BillMills:
Note continued: uploading raw data to S3 is straightforward, but mounting is a bit fraught; the instructions [here](https://github.com/s3fs-fuse/s3fs-fuse) mostly do the job, but I also needed some extra dependencies:
```
sudo yum install gcc libstdc++-devel gcc-c++ fuse fuse-devel curl-devel libxml2-devel openssl-devel mailcap
[...]

PR #122, opened 2015-12-21
First functioning Docker build

BillMills:
AutoQC now has its own functioning Docker container: https://hub.docker.com/r/billmills/autoqc/

This makes installing AutoQC and all its dependencies (including extra data) very easy (see instructions in the `README`), particularly on cloud services like AWS - setup there used to be quite complicated (see #105), but can now be done in a single script (which is also much faster than the previous setup process).

Things still to do:

 - ~~Validate dependencies. All tests pass in the docker container, but there might be some unnecessary things installed per discussion in #120~~
 - ~~Migrate travis testing. The current `travis.yml` is a bit of a mess; the Docker setup is much simpler, and guarantees we are testing and running in the *exact* same environment. Will look into moving travis testing shortly.~~
 - ~~Performance test. The internet nigh-unanimously believes that Docker is every bit as fast as native, but given the length of the full AutoQC calculation, we'd best check before we sign off on this once and for all.~~


BillMills:
Alright - about 10k profiles from *quota* ran in the docker container actually slightly *faster* than natively - probably an artefact and not a real speed-up, but good enough to convince me the claims of no performance hits are credible. Also of course, the resulting csv was character-for-character identical in both cases.

I'm going to merge this once #125 is resolved, and make sure any dependencies introduced there are reflected in the docker image. Should probably make an image `iquod/autoqc` instead of `billmills/autoqc`, too.

BillMills:
@s-good as promised, this all checks out - the only thing left to do is I'm sure you'd like the image to be called `iquod/autoqc` and not `billmills/autoqc` - if you could make a [Dockerhub](https://hub.docker.com/) account with username `iquod`, you should be able to create an empty project and give me commit rights to it, so we can push the image there and wrap this up.

BillMills:
(also notice the travis tests are way faster now, too :)

BillMills:
oh actually @s-good - it looks like there are organization accounts on dockerhub as well, so I can make one for iquod and invite you if you like, no need to make the iquod user account like I first thought.

s-good:
Hi @BillMills I'd already gone ahead and made the account but I've converted it to an organisation and have added you so hopefully all is OK.

BillMills:
No problem, it seems to have worked just fine; I'll change everything in this PR to point at the new image, triple-check it all works and we can merge - thanks, this is going to be great!

BillMills:
Alright that's got it - to recap:

 - we now have all our dependencies encapsulated in the [iquod/autoqc docker image](https://hub.docker.com/r/iquod/autoqc/)
 - this greatly speeds up and simplifies setup on Travis and AWS
 - prevents missing data problems like we wrestled with in #125 
 - ensures testing on Travis is done in the *exact* same environment as running on AWS or locally
 - guarantees we can reproduce the dependency stack on any machine at any time in the future.

Dependency details are now in `docker/Dockerfile`; this is now the 'master list' of our dependency stack, and any future dependency changes need to be reflected there.

I will be writing a wiki page on how to do fancier things on AWS with our docker container (like mounting an S3 directory with all our data into the docker container), but for now basic usage instructions can be found in the `README` - let me know any hiccups.

s-good:
Looks really nice - thanks for doing this!

PR #123, opened 2015-12-21
Introduced EN_increasing_depth_check.

s-good:
Implements the EN increasing depth check.


BillMills:
Alright, looks good - I *think* the original CI didn't download the etopo data correctly, so just restarting it fixed things; this will no longer happen when we move to a Docker image (see #122) that has the data baked right in.

Thanks for writing such thorough tests to go with the qc, looks good!

s-good:
Hi @BillMills, thanks for sorting that. I was a bit confused why it had failed!

PR #125, opened 2015-12-21
Updates to requirements with CoTeDe dependencies

s-good:
This introduces changes by @castelao to the list of requirements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ before_install:
 install:
   #- conda install -y -c https://conda.binstar.org/jjhelmus pyproj
   - pip install wodpy
-  - pip install seabird
   - pip install cotede
   - pip install gsw
   - pip install scikit-fuzzy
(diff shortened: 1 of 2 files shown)


s-good:
This build has failed like #123 because the download of data has not completed. I think this may be because curl is running in the background and the script is continuing before the download is complete. I will look into whether there is an option to change this behaviour.

BillMills:
I see what you mean, it does appear to be downloading things async. One thing to keep in mind before sinking a lot of time into fixing this: this problem will definitely go away when we start running from a Docker container ( #122 ) - it's all nominally working in my branch now, and speed tests and validation are running as I write this.

BillMills:
I'm a bit wary of the `sleep 30`. What if things take longer than 30 seconds to download?

BillMills:
there we go, `wait` is a bit more appropriate - but do please see #122; that PR is going to be ready for merge momentarily, and will change everything about `travis.yml` and `requirements.txt`, so these problems with data and dependencies will stop happening.

s-good:
Hi @BillMills, thanks for the advice - I've merged it in and am looking forward to seeing the new Docker build!

PR #126, opened 2015-12-22
Cotede

s-good:
Introduce changes made by @castelao - thanks so much for this!

One of the changes was to change the requirements.txt file to get a specific version of cotede so that the tests don't fail if cotede is upgraded. This has been lost as this file is no longer present in the latest version of AutoQC. @BillMills, what is the equivalent to this in the new version?

I think I will delete my cotede branch after this as I'm not actively keeping it up to date with the upstream repository and, if it is OK with you @castelao, please could you make pull requests to the IQuOD/AutoQC repository from now on and that way both Bill and I can look at and merge your pull requests.


BillMills:
Hi team - `requirements.txt` is now replaced with `docker/Dockerfile`. The Dockerfile is the recipe we use to re-build the Docker image when dependencies change. Also, since we are testing through Docker now as well, we can't have a `requirements.txt` at root level, since Travis will automatically assume its supposed to install this, which is no longer appropriate.

s-good:
Something's gone wrong here - the tests passed when run in the pull request but is now failing now merged into the main repository.

BillMills:
Ah, I think I see the problem - travis is pulling master when it does its checks, not the branch you actually wanted to test. I'll see what I can do right now.

castelao:
@s-good , sounds like a plan. I'll pull request directly to IQuOD/AutoQC now on.

PR #127, opened 2015-12-22
Update to allow a config to be passed direct to the cotede_test.get_q…

s-good:
…c function as well as a config file. This should correct the problem with the failing build.

### DIFF
--- a/cotede_qc/cotede_test.py
+++ b/cotede_qc/cotede_test.py
@@ -91,10 +91,14 @@ def get_qc(p, config, test):
     # Check if we need to perform the quality control.
     if p.uid() != cotede_results[0] or config != cotede_results[1]:
         inputs = DummyCNV(p)
-        with open('cotede_qc/qc_cfg/' + config + '.json') as f:
-            cotede_results = [p.uid(), 
-                              config, 
-                              ProfileQC(inputs, cfg=json.load(f))]
+        if isinstance(config, dict):
+            cfg = config
+        else:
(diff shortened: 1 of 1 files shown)


BillMills:
Note this will fail, because it is checking the failing master branch - not the patched branch you want to check. I have a fix for the build system coming momentarily, after which we can re-run this check and it ought to pass.

BillMills:
As mentioned in #128, going to merge here and hopefully that solves everything - if not, we can reassess what's still wrong.

BillMills:
Build still breaking but with different (and much fewer) errors - will attempt to fix.

s-good:
I think this may be due to the version of cotede. I'm not sure if my attempt to set the version in the docker file is working.

BillMills:
ah, great idea - remember, its not enough to just change the Dockerfile; if the deps change, we actually have to rebuild the image and push to dockerhub. I will do that now.

s-good:
Ah, OK - will do that in future. Thanks for sorting this.

BillMills:
Alright, works on my machine - new image is pushing now and will hopefully resolve all problems. Sorry for this - the janky travis.yml that started this was my fault. Transitioning to new infrastructure is bumpy at first but pays off once the dust settles.

BillMills:
Yep confirmed, that's got it; will triple check the build infrastructure right now to make sure no more funny business is waiting for us.

BillMills:
And meta-confirmed: if I break my own build, the latest travis.yml catches it properly, unlike what happened with #126.

PR #128, opened 2015-12-22
fix for travis build problems

BillMills:
currently the travis build is checking the master branch instead of the new commits it should be checking; this will hopefully fix the problem.


BillMills:
this passes on my branch but fails here, likely due to the broken build #126; going to merge here and merge #127 - hopefully this fixes everything.

PR #131, opened 2016-01-02
EN background and buddy check on standard levels

s-good:
This implements the final check from the EN suite that I plan to include in AutoQC. The others are not relevant or functionality is already included in other checks. This test is complex and the code could do with some tidying and better tests. However, results appear sensible so I am including it for now.

This pull request also includes new functionality for some tests so that they save their latest QC results to avoid repeating the calculations if other tests need to use their results. This allows the new background and buddy check to depend on the results of the other checks without enforcing a particular order for the tests to run.

The buddy check is an example of a check that requires knowledge of other profiles and this has been achieved by moving the read of profiles to the main AutoQC script so that the QC check can access it.


qctests/EN_std_lev_bkg_and_buddy_check.py
```diff
@@ -0,0 +1,239 @@
...
+    if minDist <= 400000:
+        pBuddy = profiles[iMinDist]
+        result = stdLevelData(p)
+        if result is not None: 
+            # This code should ideally be separated into a function.
+            levelsBuddy, origLevelsBuddy, assocLevelsBuddy = result
+            bgevBuddy = EN_background_check.bgevStdLevels
+            pgeBuddy      = np.ma.array(np.ndarray(len(levels)))
+            pgeBuddy.mask = True
+            for iLevel, level in enumerate(levels):
```
BillMills: @s-good this looks a little suspicious - `iLevel` enumerates `levels`, but then on the next line we use it as an index in `levelsBuddy`; is this intentional, or could this be an artefact of cutting and pasting this calculation from above (line 42)?

s-good: Hi @BillMills, sorry about this. Yes, you are right, this should be indexing levelsBuddy. As you guessed, this code is a copy and paste of some previous code. The probability of gross error (PGE) should ideally be saved in a module variable so that this calculation only has to be done once per profile.

BillMills: Great idea! Will do in #134.

Issue #133, opened 2016-01-02
EN standard level check - code tidying and testing


s-good:
An initial attempt at the EN standard level background and buddy check has been made (#131, #132). The code is complex and requires some tidying (separating code into functions etc.) Although the code appears to be doing something sensible, all the parts of the code need to have tests implemented to ensure that it is working correctly.

BillMills:
Hi Simon - happy to start having a look at this this weekend, thanks for pulling it together! I like where you're going with both the persistence of test results on the module variables `*.uid` and `*.qc` to avoid rerunning things, and on looking across the full set of profiles to find buddies. On first inspection, I have a couple of questions:

 - It seems like the results for a given test are persisted only for the current uid, and overwritten on the next, which I think means that in almost all cases, `stdLevelData` will not find a uid match and therefore will rerun the test, even though it's been run previously. I sketched out an alternative solution to this in my draft of the track check, where a global key / value store is kept for *every* profile and its qc result; so instead of having `*.uid` and `*.qc`, we could have a module variable that's a dictionary like 
```
{
  "uid_0": qc_0,
  "uid_1": qc_1,... 
}
```

 - I see you've made the profile list a global variable in #131 and examine it in the buddy check to look for buddies, without the need to a-priori sort related profiles into the same file before parallelization - I love it, that's really slick! My concern here is related to my last point; I agree your method should work, but even if we do persists a list of results for every profile, without pre-sorting it seems like those results will be in a different thread and thus invisible to their buddy, resulting in re-running tests a lot.

Does it sound like I've misunderstood something? Let me know if so. Worst case scenario is still not a showstopper - we can throw extra computing power at this to rerun some tests, but it would be nice to avoid the extra expense if possible. Let me know and I'll start tidying.

s-good:
Hi @BillMills  - thanks for looking at this! Yes, you're right on both points. As I only set it up to save the QC result from the last examined profile it is currently necessary to rerun QC for a buddy profile if it exists (this won't always happen). So it would be better to store QC results for all profiles that have been examined (as long as we don't start to eat up too much memory I guess) and then also it would be most efficient to pre-sort the profiles so that profiles that might buddy with each other are examined in the same process. I think that by having the global list it should make it so that the QC results are identical no matter how the profiles are divided up into different files, but the better the profiles are divided up the more efficient it will be, so we should get the best of both worlds.

BillMills:
[...]

PR #134, opened 2016-01-04
buddy check tune up

BillMills:
This PR will contain several updates to the buddy check and related infrastructure, per discussion in #133. Todo:

 - ~~Slice up buddy check into shorter helper functions~~
 - ~~Test buddy check helpers~~
 - Persist all qc results for `EN` tests to avoid rerunning
 - ~~Address some small issues with the buddy check:~~
  - ~~use the datastore module for passing global variables, rather than importing `__main__`~~
  - ~~currently throws a complaint:~~
```
/opt/conda/lib/python2.7/site-packages/numpy/ma/core.py:4089: UserWarning: Warning: converting a masked element to nan.
  warnings.warn("Warning: converting a masked element to nan.")
```
~~which seems ominous - should be addressed.~~

Before tackling upgrades like using the datastore and persisting qc, the first two bullets need to get done to ensure those upgrades aren't quietly mangling anything.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -1,7 +1,7 @@
 from wodpy import wod
 import glob, time
 import numpy as np
-import sys, os, json
+import sys, os, json, data.ds
 import util.main as main
 import pandas
 
@@ -79,6 +79,7 @@ if len(sys.argv)>2:
   # Identify data files and create a profile list.
   filenames = main.readInput('datafiles.json')
   profiles  = main.extractProfiles(filenames)
+  data.ds.profiles = profiles
   print('\n{} file(s) will be read containing {} profiles'.format(len(filenames), len(profiles)))
 
   # Parallel processing.
--- a/qctests/EN_std_lev_bkg_and_buddy_check.py
+++ b/qctests/EN_std_lev_bkg_and_buddy_check.py
@@ -116,12 +116,15 @@ def buddyCovariance(meso_ev_a, meso_ev_b, syn_ev_a, syn_ev_b):
     meso_ev_a == mesoscale error variance for profile a, etc.
     '''
 
-    corScaleA = 300.0 # In km.             
+    corScaleA = 100.0 # In km.             
     corScaleB = 400.0 # In km.
-    corScaleT = 21600.0 # In secs.
+    corScaleT = 432000.0 # In secs.
     mesSDist  = minDist / (1000.0 * corScaleA)
     synSDist  = minDist / (1000.0 * corScaleB)
     timeDiff2 = (timeDiff(profile, buddyProfile) / corScaleT)**2 
+
+    if timeDiff2 is None:
+        return None
 
     covar = (meso_ev_a * meso_ev_b *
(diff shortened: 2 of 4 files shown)


BillMills:
I noticed that the correlation scales used in `buddyCovariance` do not match those given in the text; we have a 6 hour correlation time, while they use 5 days, and we use a 300 km meso correlation length, while they have 100 km. I have no opinion - just want to make sure any departures from the text are intentional and documented.

Also, I see an interesting case in `timeDiff` - if the day is 0, it's set to 15. I imagine the idea here is that the day should never be 0, so 15 is the most agnostic choice in a 30 day month; again, happy to go with this if it's what we want, but I'd be given to return `None` or some other indication that there isn't enough information to do this calculation in this case.

s-good:
Hi @BillMills, thanks for being so thorough about checking this!

You are correct on all these things: meso correlation length should be 100 km, time correlation should be 432000.0 (5 days) and I think you are right that the day shouldn't be set to 15 if it is not available and the calculation shouldn't be done.

BillMills:
Just noticed something novel [here](https://github.com/IQuOD/AutoQC/pull/131/files#diff-aa4801db17bc02d927c483ba27f406dcR183) in the original implementation of this check; using `origLevel` instead of `i` to index `preQC` would usually throw an error - except that `origLevel` is always a bool, which Python typecasts to 0 or 1 as an index, thus returning something sensible but probably unintended. The latest commit changes this to indexing by the enumerated index; if this was actually right the first time and I just don't get it, make sure to let me know!

s-good:
I may have messed up how this is working but origLevels shouldn't be boolean. The data saved in EN_background_check is a subset of the original profile and origLevels is supposed to record, for each of the saved levels, the index of the level in the original profile.

BillMills:
@s-good right you are, I think I got `origLevel` confused with `preQC` - you had it right the first time.

Anyway, this is my best first pass at slicing up the buddy check and unit testing what I could capture in meaningful functions; I'll leave it to you to give it the once over before we merge. Some chunks are still undertested - I'm looking at the main `test`, and `update_pgeData` - but I'm not sure pulling those apart into any smaller bits will make sense; it'd be good to have a profile we know exactly what to expect from, and do an integration test of the whole thing to wrap up here.

s-good:
Hi @BillMills, it looks good - thanks so much for doing this! What was the cause of the warning about converting the masked element to a nan?

In the future I will try to find some real profiles to add to the tests for this QC check. I have opened a new issue and assigned myself to remind me to do this (#136).

BillMills:
> What was the cause of the warning about converting the masked element to a nan?

To be honest it just kind of went away - my best guess is that in the pge calculation, we were originally conflating the primary profile's pge with the buddy's, a cut-and-paste bug we discussed earlier; this could have resulted in choosing valid levels to look at from one mask, and finding masked elements (which subsequently got converted into nans) from the other.

Maybe? It went away somewhere in a09be4a to 6080ab3, where the most substantial change was straightening out the above.

Anyway, I think we can call it a day here, modulo the other issue you opened - thanks!

PR #135, opened 2016-01-16
Added the first ICDC test.

s-good:
Implemented the ICDC level order test.

This includes functionality to retrieve the reordered profile depth and temperature information for use by other tests and to reorder data back into the original order.


qctests/ICDC_aqc_01_level_order.py
```diff
@@ -0,0 +1,146 @@
...
+    # again.
+    llpos      = zr >= 0.0
+    zr         = zr[llpos]
+    tr         = tr[llpos]
+    origlevels = origlevels[llpos]
+    nlevels    = np.count_nonzero(llpos)
+
+    return None
+
+def order(t, n):
```
BillMills: Are there builtins in numpy (or just plain `.sort()` or `sorted()`) that will achieve the same effect? Builtins are usually fastest.

s-good:
Hi @BillMills, yes you are probably right. I was going down the route of making the python version as similar as possible to the fortran but maybe that's the wrong approach. I'll take another look.

BillMills:
That's a good place to start from! Fortran is super quick no matter what you do, but python loops can sometimes slow you down (the builtins are probably fortran under the hood :).

s-good:
The code has been reworked as suggested. Results match the test profiles supplied with the fortran so I will merge this pull request and close.

Issue #141, opened 2016-01-27
Bug on DummyCNV. Only affects CoTeDe tests


castelao:
It was missing a round to define the seconds at DummyCNV. Approximately 50% of the cases would have a 1 second off.

This only affects CoTeDe tests, and it is just a conceptual error. It shouldn't effectively change any result.

s-good:
Hi @castelao, thanks for this and also for your work adding the Wod4CoTeDe class to wodpy. I guess this means that we can eventually remove the DummyCNV class and use the Wod4CoTeDe class instead?

castelao:
Hi @s-good, yes, that's the plan. But don't worry about this, you two already have plenty on your plate. I'll be pushing updates as I can move on.

s-good:
Hi @castelao, something just occurred to me - will the rounding of the number of seconds work if seconds >= 59.5? Will this get rounded up to 60, which isn't a valid number of second?

castelao:
Good call @s-good 

Let me try again:
https://github.com/castelao/wodpy/commit/4bd8d0e564fc987ea97c822b17d75d405c4bc189

s-good:
This solution looks really good. A lot more elegant than manually calculating the hours, minutes and seconds!

s-good:
Just thought, you will need to replace the hours = 0, minutes = 0 and seconds = 0 lines in the if statement with time = 0 or similar.

castelao:
Yes, I realized just after I submitted that. 
https://github.com/IQuOD/wodpy/pull/5/files

Thanks!

BillMills:
Good to close here?

PR #142, opened 2016-01-27
BUGFIX #141, missing a round to define the seconds.

castelao:
Minor bugfix. It shouldn't compromise the previous results.


s-good:
Thanks for picking this up!

PR #147, opened 2016-02-19
Added ICDC local climatology test.

s-good:
This pull request adds the last ICDC test. On the small 150 profile test dataset this achieves an impressive 83% true positive rate, although at a cost of 14% false positives! It will be interesting how this translates to the full test dataset.

The test relies on a large (> 2GB unzipped) data file. I need to get this made available on an ftp site. The read of the data from the ASCII file is very slow so on first access there is a conversion to a netCDF file which is far, far faster to read. The netCDF is then read on subsequent occasions.

Tests will currently fail because the data file is not available. I will update this pull request once it is available.


BillMills:
Thanks, Simon! Would it help to just store the parameter file as netCDF to begin with?

s-good:
Hi Bill, OK, I've tweaked the requirements for the test so that it does not require the ASCII data file if the netCDF is present.

s-good:
I have updated the docker image with the data file so I think that tests should now pass. I'll push a commit in a moment and see if it works.

s-good:
All looks fine so I will merge this pull request.

PR #148, opened 2016-02-23
Fixed typo in README.md

PAVBAN95:
"each" found two times in a row, and Matlab should be written as MATLAB, and "ie" is replaced by "i.e."


BillMills:
Thanks, Pavan!

PAVBAN95:
Your welcome, BillMills!

PR #151, opened 2016-03-23
Optimize CoTeDe

castelao:
pypi is having problems with its indexes, so an alternative to keep the packages updated might be to `pip install --upgrade` when generating the docker.

If that works, the common flags at CoTeDe will not be required anymore.


castelao:
Found a better solution. I'm canceling this.

Issue #136, opened 2016-01-17
EN standard level check - add tests using real profiles


s-good:
The testing of the EN standard level check would benefit from some real world examples to confirm it is working as expected.

PR #155, opened 2016-03-29
EN standard level check - real tests, bug fixes and final part

s-good:
This pull request adds validation tests based on real profiles for the EN standard level checks (#136). In doing this I found a bug that affects both the standard level check and the background check and I also fixed a couple of other bugs I found in the standard level check. Also included is the final bit of functionality for the standard level check, which reinstates levels if they are consistent with the levels above or below.


BillMills:
r+ thanks @s-good!

PR #156, opened 2016-03-29
Fixes for runtime errors when running EN track check.

s-good:
This pull request fixes a couple of issues I found when running the EN track check on the full QuOTA dataset:

- util/geo.py needs to handle missing data when creating datetime instances.
- The test crashes if number of data files > number of threads because the EN_track_headers variable is not {} after the first file is processed in the thread. I've attempted to fix this by saving the data file name and updating EN_track_headers when this changes.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -36,6 +36,7 @@ def processFile(fName):
 
   # keep a list of only the profiles in this thread
   data.ds.threadProfiles = main.extractProfiles([fName])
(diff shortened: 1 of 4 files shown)


BillMills:
@s-good good catch, I didn't consider *not* parallelizing over multiple files. It would be interesting to see if results are globally the same for two files on one thread vs. the same files on separate threads - but that's a task for another day. Thanks!

PR #159, opened 2016-04-02
Update CoTeDe

castelao:
pypi is not working well, so let's use the releases from github.

@BillMills and @s-good, how would be the best way to define an environment variable inside the docker?
 ``OCEANSDB_DIR='/AutoQC/data/'``


BillMills:
Hi @castelao,

Sounds good - re: env variables in the Dockerfile, that's something that needs to get sourced every time the container is started - which sounds like something to put in a `.bashrc` script. After writing a `.bashrc` script, you can include it in the Dockerfile with an `ADD` command, like the datafiles - also, the `.bashrc` should be tiny, so feel free to add that to the `/docker` directory in this repo. Once that's there, I can update the Docker image with that script in place.

castelao:
Hi @BillMills,

done.

BillMills:
Hi @castelao,

When I rebuild the docker image using the Dockerfile in this PR and run the tests, I get the following errorlog: 
[new_docker_image_errorlog.txt](https://github.com/IQuOD/AutoQC/files/201910/new_docker_image_errorlog.txt)

I will look more carefully at these errors as soon as I can - meanwhile, can you build the docker image from your Dockerfile locally and see if you get the same errors? Thanks!

castelao:
Hi @BillMills,

I believe I addressed all errors. Could you confirm it, please?

The actual performance improvements will be on the following PR.

BillMills:
That's progress - there is one less fail and one less error now: 
[new_docker_image_errorlog_2.txt](https://github.com/IQuOD/AutoQC/files/209545/new_docker_image_errorlog_2.txt)

All the errors I'm seeing make the complaint:

```
UnboundLocalError: local variable 'g' referenced before assignment
-------------------- >> begin captured stdout << ---------------------
Deprecated cfg format. It should contain a threshold item.
Deprecated cfg format. It should contain a threshold item.

--------------------- >> end captured stdout << ----------------------
```

There were there last time, too - could there be a version of a dependency on your machine that is not being reflected in the Dockerfile?

castelao:
It's probably because pip is not upgrading the versions already in the docker (wodpy, OceansDB, CoTeDe).

One solution is to:
`RUN pip install --upgrade git+https://github.com/...`
but that would update all dependencies, including numpy, scipy, etc. Would that be a problem?

BillMills:
`pip freeze` inside my docker container as it is built looks like:

```
conda==3.19.0
conda-env==2.4.5
cotede==0.17.0
cycler==0.9.0
filelock==2.0.5
gsw==3.0.3
matplotlib==1.5.0
netCDF4==1.2.1
nose==1.3.7
numpy==1.10.2
oceansdb==0.5.0
pandas==0.17.1
pycosat==0.6.1
pycrypto==2.6.1
pyparsing==2.0.3
pyproj==1.9.4
python-dateutil==2.4.2
pytz==2015.7
PyYAML==3.11
requests==2.9.0
scipy==0.16.1
seabird==0.7.0
Shapely==1.5.13
six==1.10.0
wheel==0.26.0
wodpy==1.3.0
```

You're right that Docker won't update a line like 
```
RUN pip install git+https://github.com/castelao/oceansdb.git@master#egg=oceansdb
```

just because someone pushed to GitHub - it'll look to Docker like nothing has changed, and will use its cached version of that command's result. I will try busting the cache and seeing if that makes a difference.

castelao:
CoTeDe 0.17.0 is the problem, it should be >= 0.17.2

BillMills:
yep, all tests are passing for me now - the new image is uploading to Dockerhub as we speak, which should pass the tests on Travis and allow us to merge.

castelao:
Terrific! Once that's ready, I'll submit the actual performance improvements.

BillMills:
OK, looks good - the only (non-blocking) thing left, is some warnings getting raised:
```
/opt/conda/lib/python2.7/site-packages/numpy/ma/core.py:3114: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future

  dout = self.data[indx]

/opt/conda/lib/python2.7/site-packages/numpy/ma/core.py:3169: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future

  dout._mask = _mask[indx]
```

Sounds like this may break in a future numpy version - I suppose we can fix it then, but if the problem is obvious it might be worth patching now.

PR #162, opened 2016-04-28
Change Argo tests to work on pressure instead of depth.

s-good:
Argo tests are designed to work on pressure so this change corrects the previous implementation. GSW (Python version of the TEOS-10 software) is now used for depth-pressure conversion.


BillMills:
Hi @s-good - sorry for the absurdly slow review here, been traveling on a couple other projects, but I'm back! This looks great, thanks for handling this.

PR #169, opened 2016-06-08
stopping wod test from breaking when datafiles.json changes

BillMills:
while all tests are passing happily on travis, nose kept killing my test suite locally - this was because I had been running tests after fooling around with `datafiles.json`, which `wod_test.py` was trying to read. This PR decouples the two.


BillMills:
travis isn't picking this PR up, going to try closing and reopening.

PR #172, opened 2016-06-12
Cotede config format

BillMills:
At the moment, running the CoTeDe tests produces a huge number of warnings:

```
Deprecated cfg format. It should contain a threshold item.
``` 

This is due to a change in config file format in CoTeDe, which this PR attempts to address. The json changes in `cotede_qc/qc_cfg/` are meant to reflect the pattern and spirit of https://github.com/castelao/CoTeDe/commit/23dfdca96da7942514d4563da3fa26f23d5620a5 and https://github.com/castelao/CoTeDe/commit/3819361becc7669542aadc0892ae2d653ad06089, and have the desired effect of eliminating the warnings - but as I have not had much involvement in the CoTeDe integration part of this project, I'd like a second opinion before we merge; cc @castelao @s-good.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -39,8 +39,6 @@ def processFile(fName):
   data.ds.threadFile     = fName
 
   for iprofile, pinfo in enumerate(data.ds.threadProfiles):
-    if iprofile >=10:
-      continue
     # Load the profile data.
     p, currentFile, f = main.profileData(pinfo, currentFile, f)
     # Check that there are temperature data in the profile, otherwise skip.
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -3,7 +3,7 @@ import glob, time
 import numpy as np
 import sys, os, json, data.ds
 import util.main as main
-import pandas
+import pandas, logging
 
 def run(test, profiles):
   '''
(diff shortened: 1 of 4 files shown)


s-good:
This certainly looks reasonable to me. These config files were taken from CoTeDe and then trimmed to remove the bits not needed for AutoQC, so it is correct that any changes to the CoTeDe config files should be matched in AutoQC versions.

s-good:
I've just been looking at #173 and have realised that the code that runs the CoTeDe tests has changed considerably since I last worked on it. It looks like these config files may now be not needed. The warnings are probably being triggered by the fuzzy logic config file since this config is now called fuzzylogic in CoTeDe instead of fuzzy. However, #173 reveals that there is an issue that things are not working quite right. I will reply to that issue to explain.

s-good:
As commented in #173 and #176 the CoTeDe copies of the config files were not being used due to a missing import and other problems. This was fixed in #175. Once #176 is complete the AutoQC copies of the config files will no longer be needed. The warning messages described above should not appear once https://github.com/castelao/CoTeDe/pull/29 is merged.

BillMills:
Sounds right to me - let's close this PR out in favor of the stack of changes you cite above.

PR #180, opened 2016-08-01
Sqlite functionality and EN std level check working with db

s-good:
Here is my attempt to get the EN std level check working with the new database structure. Tests pass and results on the quota_subset.dat dataset appear to be identical to before.

I was struggling to get the postgres to work outside the docker container so I've added in a fallback to use an sqlite database if postgres is not available. Hope this is OK. It seems to work with almost the same commands as postgres. Main difference is that it is necessary to store qc results as integers as booleans can't be stored, but I thought that would be OK. 

I think I messed up when I started work on this as I seem to have got all the recent commits to the master branch mixed up with this one - sorry about that!


BillMills:
Thanks for this, @s-good! I'm a bit nervous about using sqlite for autoqc though, since it has concurrency difficulties that postgres avoids, which is why I chose postgres; see http://sqlite.org/faq.html#q5 for example. The level of parallelization we need in order for these calculations to not take weeks means that concurrent db access is a key priority.

Could I help either (a) getting postgres working on your machine or (b) getting running inside the docker container working for you? That helps us avoid a lot of complicated cases as well. In either case, I think the changes you made to the level check should be mergable on their own; I have most of Tuesday set aside to work on this, so I'll do some git surgery to pull those changes in then.

s-good:
Hi @BillMills, I think that the postgres will work fine at home running in docker but if I'm running it at work I can't use docker and I'm going to struggle a lot with postgres because of having to set up a server. It sounds like sqlite is ok as long as using an ext4 file system and not too many threads so I'd like to include it to enable me to try to test running at work. NB I realised that what I have set up with the try: import psycopg2 will not work and there will need to be a command line argument instead.

BillMills:
@s-good okay, we can make sqlite work - but actually, I would rather just drop postgres and use one single database solution to avoid unnecessary complexity in that case. What do you think? If this is acceptable, let me know and I will make all the necessary changes immediately.

s-good:
Having been playing with it a bit today I'm not sure if it is wise to put all our eggs in sqlite basket. I seem to be getting different behaviour from it depending on where I run it which makes me very nervous and it sounds like postgres would be better in general. I can make local modifications if necessary when running at work if you want to remove the sqlite parts. It is straightforward to do as commands are almost identical. I can run using docker when running it elsewhere and then there is no issue with postgres.

BillMills:
@s-good okay, let's do that - I will try and turn this around by the end of Wednesday, giving us a couple more days to validate at scale this week; if nothing goes wrong there, we can move onto large scale running by the weekend or first thing next week.

BillMills:
I'm going to close this as I have pulled the key changes for the standard level check into `postgres`, in a postgres-only scheme. @s-good, I went back to using bools for test results, but we don't have to; we can use 0 for fail / no problems and >0 for flags if that's convenient (also makes it easier in future to have different flavors of flags).

Issue #165, opened 2016-05-30
Profiles pre 1900


BillMills:
Hi @castelao - CoTeDe tests are currently producing errors on profiles from years earlier than 1900:
```
ValueError: year=1773 is before 1900; the datetime strftime() methods require year >= 1900
```
before I go hacking on these, have you solved this previously for CoTeDe? Let me know if you have a preferred solution.

BillMills:
At the moment when we find a profile pre 1900, we're just fudging the year to be 1900 and going ahead with running; remember to revert this when this gets fixed upstream.

PR #182, opened 2016-08-05
Exclude profiles pre-1900

s-good:
As mentioned in #165 CoTeDe uses a datetime function that cannot handle dates before 1900. This filters out those observations from being put in the database.

Alternative is to leave them in and exclude them from going through the quality control in AutoQC.py. However that could cause confusion later when analysing results and it is straightforward to add extra rows to the database at a later time if we want to process them.

NB this only affects a very small number of profiles - in QuOTA is ~180 out 350000

### DIFF
--- a/build-db.py
+++ b/build-db.py
@@ -56,9 +56,16 @@ if len(sys.argv) == 3:
         wodDict['raw'] = "'" + raw + "'"
         # Below avoids failures if all profile data are missing.
         # We have no use for this profile in that case so skip it.
+        donotuse = False
         try:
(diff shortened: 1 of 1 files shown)


BillMills:
What if we ran on everything, and made CoTeDe automatically pass everything pre 20th century? That way we avoid nulls and having to run multiple times - unless we're just not that interested in this early data.

s-good:
That's another option. I was thinking something simple as it's such a small number of profiles. We could also set the year to 1900 for pre-1900 profiles, which would allow the tests to run. I don't think there is any impact of this on the qc results. I'll close this and open a new one with that instead.

PR #183, opened 2016-08-05
Update cotede_test.py to handle pre-1900 profiles

s-good:
woa_normbias can't handle pre-1900 profiles because it uses datetime.datetime.strftime which fails if given a date before 1900. Profiles from before 1900 are set to have year=1900. This should not have an impact on QC results but will prevent an error.


s-good:
I think that the update to CoTeDe must have changed the QC decisions. It's affecting the postgres branch as well, although there is an additional error message in this branch. I'll investigate as soon as possible.

s-good:
I've updated the expected QC decisions. Only failure is a complaint about lack of 'threshold' being explicitly stated in the config. This would be fixed by https://github.com/castelao/CoTeDe/pull/29 but as far as I know does not impact on results.

BillMills:
Looks like this is passing on iquod/autoqc/postgres - thanks, @s-good!

Issue #190, opened 2016-09-02
Global variables


s-good:
Hi @BillMills, I've been thinking a little bit about what you told me about global module variables being a problem when running with multi processors. This is awkward as some of the EN checks (particularly EN std level check) relies on this. I wonder if a possible work around is to remove the parallel running from within the Python code and set it up so that we can tell it to process a particular set of profiles in serial. E.g. we would run it like so:

`python AutoQC.py test 0 1000`

Where the 0 and 1000 tells it to process profiles 0 through 1000. Then to get the parallel processing we could have a shell script or another Python program that automatically spawns lots of the AutoQC processes, each processing a different set of profiles.

I think this will avoid the global variables issue as each AutoQC process has its own set of global variables and will be processing the profiles one at a time, but we will still get the speed up from running in parallel. What do you think?

BillMills:
I like this idea and wish it would work, but if I've understood correctly (possibly not), Python's [global interpreter lock](http://stackoverflow.com/questions/1294382/what-is-a-global-interpreter-lock-gil) might get in the way of this - a whole bunch of python programs running simultaneously actually take turns in a single queue. 

Luckily though, I have almost finished solving the problem for the EN tests on branch `no-global`. Basically, I'm just:

 - checking the database for precomputed QC results instead of having them hang out on a global variable
 - parking the other globals on the `parameters` dictionary that gets handed around between all the tests. This works essentially like a global data store, but avoids confusing scope issues that the old `data.ds` datastore might have had, being imported separately in a bunch of threads. I think this will work, though the current implementation on `no-global` still needs to be 'namespaced' by uid, so concurrent profiles don't overwrite each other's 'globals' - but I think I can get this done tonight.

All told, I'm hoping `no-global` will be complete, validated and mergable by EOB Monday.

s-good:
I think the GIL only applies if using threads (see https://docs.python.org/2/library/multiprocessing.html which says that it avoids the GIL by using sub-processes rather than threads). This would be completely separate Python processes so shouldn't be affected (although I should caveat this by saying that I don't really know what I'm talking about here!)

Have you found a way to store the data saved in the EN background check, which is then used in the EN std level check (and also I think in the EN background available check)? My concern with going down the route of accessing results from the database and/or saving EN background check data in memory is the potential to slow everything down through disk access (either by have to access the database file or by filling up physical memory so that we have to start using virtual memory). 

[...]

PR #191, opened 2016-09-05
Process-level parallelization

BillMills:
this PR implements @s-good's suggestion for parallelization from #190.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -57,7 +57,9 @@ if len(sys.argv)>2:
       return
 
     # run tests
+    print uid
     for itest, test in enumerate(testNames):
+      print test
       result = run(test, [profile], parameterStore)
       query = "UPDATE " + sys.argv[1] + " SET " + test.lower() + " = " + str(result[0][0]) + " WHERE uid = " + str(profile.uid()) + ";"
       main.dbinteract(query)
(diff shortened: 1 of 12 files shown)


BillMills:
Trying to sort out why this build is failing:

 - tests all pass in the docker container locally and on AWS
 - tests fail on Travis after attempting to re-download parameter files needed for CoTeDe [like this one](http://data.nodc.noaa.gov/thredds/fileServer/woa/WOA13/DATAv2/temperature/netcdf/decav/5deg/woa13_decav_t13_5dv2.nc).
 - Trying to navigate to the parameter files in a browser returns an error at the time of this writing, with NOAA reporting their servers are down.
 - Re-running [Travis build 315](https://travis-ci.org/IQuOD/AutoQC/builds/155767077) is failing this same way currently, even though it passed previously.

All of which adds up to what seems like Travis is unnecessarily trying to re-download these parameter files and may have been happily doing so for some time, until tonight when those links to NOAA ceased to function. None of which should be necessary, as the relevant parameter files are baked into the Docker image, and work perfectly everywhere but Travis. @s-good, can you point out where we tell CoTeDe to go looking for the parameter files before it tries to download them? If this is somehow breaking on Travis, it might explain this behavior.

PR #193, opened 2016-09-06
Process-level parallelization II

BillMills:
This PR replaces #191, with the key update that bashrc files are not run when not logging into a docker container interactively, which is the case on Travis; therefore, environment variables must be set in the Dockerfile. Note the new Docker image hasn't been pushed at the time of this writing (so tests will still be failing), but should become available in about an hour.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,8 @@ services:
 before_install:
 - docker pull iquod/autoqc
 
+env:
+- OCEANSDB_DIR=/AutoQC/data/
+
 script:
 - docker run -v $PWD:/AutoQC_latest iquod/autoqc /bin/bash -c "cp /AutoQC/data/* /AutoQC_latest/data/.; cd /AutoQC_latest; nosetests tests/*.py"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,5 @@ services:
 before_install:
 - docker pull iquod/autoqc
 
-env:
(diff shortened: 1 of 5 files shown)


BillMills:
Alright as promised, that's got it; @s-good, feel free to merge this PR if you're satisfied, and go ahead and start pushing our data through en masse. A few comments:

 - see `launchQC.sh` for the process-level parallelization, comments are hopefully clear.
 - Once you're completely satisfied that the code + docker image is what we're using for v1, we should tag a release of both so that we can always reproduce this moment in time easily.
 - I would strongly recommend running everything except the EN suite first, as everything else will finish quickly and give us the bulk of our info; then we can leave the EN suite grinding away for the rest of the time while we play with the other results.
 - We'll need some way to capture the postgres table after all's done, to feed those results into whatever final decision making process we settle on; I suspect persisting the actual postgres table shouldn't be so difficult, but dumping it all to a great big CSV is certainly easy to do in a pinch.
 - Let me know if funding materialized and you'd like to throw some cloud resources at the production run.

s-good:
Hi, looks good. Just to check on the launchQC.sh is the call sequence

bash launchQC.sh <database table name> <number of profiles> <number of processes>
(it doesn't say the database table name argument in the first line of the script but looks like it is needed).

Also in launchQC.sh, in the final line should it say $1 instead of demo?

BillMills:
Ah - yes to both!

s-good:
Looks great!

Issue #209, opened 2017-09-22
SQRT of negative numbers issue


s-good:
This line:

https://github.com/IQuOD/AutoQC/blob/1719118337a3d02e1847a6de9f402d2db182c5ca/util/geo.py#L18

sometimes causes a failure as hs sometime can be negative. It needs a check for a negative number and sqrths set to 0 in that case.

BillMills:
Excellent catch, and in a surprise twist this seems to be the root cause of some regressions between our postgres and sqlite implementations I've been chasing around. But, should we set it to 0 when the argument is negative, or should we take the absolute value of the argument? Haversines should always be positive and safe to square root. If a negative value is appearing, I'm worried that something more pathological is happening deeper in.

BillMills:
For example, EN_track encounters the following three lat / long pairs:

-36.1669, 20.0169
-34.8669, 20.0169
-34.9, 20.0169

Representing a craft that *exactly* about-faced. Our implementation of `haversineAngle` (at least on my machine) comes up with `haversine(c) - haversine(a-b) == -5.42101086243e-20`, creating a negative argument in its subsequent use of `arcHaversine`, by what looks a lot like a rounding error.

In addition to this behavior, close inspection of the same functions (and assuming these rounding errors are dealt with) reveal that in the case above, where I would have expected the angle to be 180, we'll get back 0 degrees from `haversineAngle` as `math.asin(0)` in `arcHaversine` returns 0, expecting us to deal with the 180 degree modulo by our own devices. This will cause systematic errors in `EN_track` when it fails to distinguish dead ahead from about face.

I'll rework our haversine functions soon to address both these problems.

PR #210, opened 2017-09-27
correct and protect course correction angle calculation

BillMills:
at least a partial response to the discussion in #209:

 - Avoids the use of `archaversine` when calculating angles between track segments, and the corresponding pathology, by using the analogous spherical law of cosines instead, with enforcement of [-1,1] arguments for arccos.
 - Doesn't actually do anything about archaversine itself; as I said in #209, I think that function should complain about negative arguments, since (non-imaginary) haversines should always be positive.


BillMills:
thanks for checking, @s-good; after this correction, I get the exact same results between our postgres and sqlite implementations in a single process; will try a multi-process run and let you know.

PR #213, opened 2017-12-30
protect against duplicate profiles when building db

BillMills:
Now that we use uid as a formal primary key when building our initial database in `build-db.py`, we must deal with the situation where a duplicate profile appears in the raw wod-ascii data (as one does in my copy of the full quota set). Here we keep track of what uids have been previously written to the database, and skip any subsequent repeats.


BillMills:
@s-good sounds good, now reports duplicate UID, file it was found in and position in the file of the offending profile.

PR #215, opened 2018-01-13
Correcting the markdown in the README.md

moijes12:
The markdown in the below sections of the README.md were not having the desired effect:
 * Introduction
 * Usage 
 * Testing

Corrected the sections so that the markdown has the desired effect


BillMills:
Thanks, @moijes12!

PR #217, opened 2018-01-25
Handling temperature 99 error codes

BillMills:
It seems that 99.99, in addition to 99.9, has been used as a mask; see UIDs 66424634, 66424635, 66424645, 66425567, 66425571 for example. The UID's mentioned make up about half the false positives in the optimal point of the ROC curve I'm currently looking at, so this seems impactful; @s-good let me know if a temperature of 99.99 is indeed to be masked as I assume, or if it means something more nuanced.

### DIFF
--- a/util/main.py
+++ b/util/main.py
@@ -19,7 +19,7 @@ def importQC(dir):
 
 def catchFlags(profile):
   '''
-  In some IQuOD datasets temperature values of 99.9 are special values to
+  In some IQuOD datasets temperature values of 99.9 or 99.99 are special values to
   signify not to use the data value. These are flagged here so they are not
   sent to the quality control programs for testing.
   '''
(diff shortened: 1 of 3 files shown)


BillMills:
Update: on the topic of temperatures of 99.9 and 99.99, I now notice that while `main.catchFlags` was being used to scrub these out before passing profiles to our QC tests, nothing seems to have been done in the post-analysis to do similar accounting, such as for example `analyze-results.py` to build our ROC curves; @s-good please correct me if I'm wrong.

Today's commits rectify this by setting the QC code to 99 for every level that reports a temperature between 99 and 100 degrees (to accommodate the several different 99.xx codes I point out above) when building the original database in `build-db.py`. This has the advantage of being as nondisruptive as possible; `dbutils/summarize_truth` can omit these without any further refactoring. The drawback is that it overwrites whatever the original QC code was for this level - but seeing as we never consider QC codes beyond whether or not they're 3 or greater, this doesn't really hurt us.

util/main.py
```diff
@@ -28,7 +28,7 @@ def catchFlags(profile):
   for i in range(profile.n_levels()):
       if profile.profile_data[i]['variables'][index]['Missing']:
           continue
-      if profile.profile_data[i]['variables'][index]['Value'] == 99.9:
+      if profile.profile_data[i]['variables'][index]['Value'] >= 99.9 and profile.profile_data[i]['variables'][index]['Value'] < 100:
```
s-good: Should this be >=99?

s-good:
I think it is a good idea to make this change. I think that the assumption was that if the temperature was 99.9, the 'truth' QC flag was not set to reject. I don't think this is a safe assumption, though.

PR #220, opened 2018-02-14
improve performance of EN_increasing_depth

BillMills:
The current implementation of `EN_increasing_depth` scales very badly (as bad as n^3) in number of profile levels, due to the recomputation of the entire nxn matrix of level comparisons, as many as n times as each level is examined. This causes profiles with a large number of levels (see uid 86124230 with ~8000 levels) to take an enormous amount of time.

If I've understood the algorithm correctly, it shouldn't be necessary to recompute the entire matrix at every iteration; when a level becomes flagged by this test, it should only be necessary to recompute the corresponding row and column of the comparison matrix, dropping the execution time to n^2 instead of n^3. This PR implements that change; @s-good please examine critically as this is a non-trivial refactor of this test.

### DIFF
--- a/qctests/EN_increasing_depth_check.py
+++ b/qctests/EN_increasing_depth_check.py
@@ -6,7 +6,6 @@ import EN_spike_and_step_check
 import numpy as np
 from collections import Counter
 import util.main as main
-import sys
 
 def test(p, parameters):
     """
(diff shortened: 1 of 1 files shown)


BillMills:
even better! done per your suggestion.

PR #221, opened 2018-03-02
more thorough wire break test

BillMills:
It occurs to me that `util.dbutils.db_to_df`'s `filter_on_wire_break` option seems to assume that `CSIRO_wire_break` test flags all levels that are affected by a wire break; I'm not sure that it does. Most profiles flagged by the wire break test look qualitatively like:

But [the wire break reference](https://journals.ametsoc.org/doi/full/10.1175/JTECHO539.1) and our qc test only check the very last level - meaning that the db util is only filtering out, at most, the last level of a profile. Is that really what we want, or do we want to filter out all the red points in the above figure? (the unlabeled color scale is originator qc code).

This PR tries to flag all such points at the end of a profile by flagging:
 - all the levels in a consecutive run from the end that fall outside of -2.8<t<36
 - all consecutive levels shallower than this run where the gradient from the next-shallower level is > 0.5 deg C / m in magnitude.

This is somewhat arbitrarily constructed - @s-good please let me know if I've misread something entirely, and also how we can make this a little more precise. Also before we merge, I'm going to run this over a few 10's of thousands of profiles just to check for edge cases.

### DIFF
--- a/qctests/CSIRO_wire_break.py
+++ b/qctests/CSIRO_wire_break.py
@@ -1,9 +1,16 @@
 """
-Implements the wire break test of DOI: 10.1175/JTECHO539.1
-All questionable features result in a flag, in order to minimize false negatives 
+Implements the wire break test of https://github.com/BecCowley/Mquest/blob/083b9a3dc7ec9076705aca0e90bcb500d241be03/GUI/detectwirebreak.m
 """
 
 import numpy
+
+def istight(t, thresh=0.1):
+    # given a temperature profile, return an array of bools
+    # true = this level is within thresh of both its neighbors
+    gaps = numpy.absolute(numpy.diff(t))
+    left = numpy.append(gaps,0)
+    right = numpy.insert(gaps,0,0)
+    return (left<thresh) & (right<thresh)
 
 def test(p, parameters):
     """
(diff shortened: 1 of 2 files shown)


BillMills:
Right! What do you think, @BecCowley? Is it fair to call all the red points above symptomatic of a wire break, or is that too heavy-handed? If these are all part of a wire break event, do you have a preferred algorithm that flags them all? I'd prefer to use something established if it exists, rather than the rough estimate I propose here.

BecCowley:
@BillMills, all the red points are wire break points and should be flagged as bad. You will need some sort of gradient or difference test to get all the points. Then anything deeper than the point is bad.

Here is a link to our automatic wire break detection code used in our XBT QC gui (Mquest):

https://github.com/BecCowley/Mquest/blob/083b9a3dc7ec9076705aca0e90bcb500d241be03/GUI/detectwirebreak.m

You can see it has a range detection and a difference cutoff between points. This works really well in probably 90% of our high resolution data.

BillMills:
Sorry this took a while, I've been 100% focused on the track check for the past while. Anyway - this now reproduces Bec's suggestion linked above.

PR #222, opened 2018-03-11
precompute pressures in en_background

BillMills:
By transforming depths to pressures for every level in the profile all at once in `en_background`, we can take better advantage of `gsw`'s optimizations (it's really quite well written). Try the following sketch to see an illustration of this:

```
import gsw, numpy
from timeit import default_timer as timer

d = numpy.array(range(2000))
lat = 0

start = timer()
for z in d:
    gsw.p_from_z(-z, lat)
end = timer()
print end-start


start = timer()
gsw.p_from_z(-d, lat)
end = timer()
print end-start
```

As usual, @s-good let me know what you think.

### DIFF
--- a/qctests/EN_background_check.py
+++ b/qctests/EN_background_check.py
@@ -3,7 +3,7 @@ Implements the background check on reported levels from the EN quality control
 system, http://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf
 """
 
-import EN_spike_and_step_check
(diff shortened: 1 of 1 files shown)


BillMills:
@s-good a wise precaution, I agree. I updated this to replace missing depths with 30000 before computing pressures, leaving them doubly filtered: once by the `isData` check in the subsequent loop, and once by the 20k limit on pressure in `obs_utils.pottem`.

PR #225, opened 2018-05-15
Best QC set using at least one from each type of test

s-good:
This pull request is an attempt to implement finding a best QC set by enforcing the use of at least one from each type of QC test that is designed to catch particular error modes (such as a spike). Updates are:

- Read QC group definition from a CSV file; this allows filtering the data using the results of particular tests.
- Five QC tests were 'pre-chosen': wire breaks test, checks that location and time data are plausible, check that depth is zero or greater, and reject of XBT levels < 3.6 m (which is a standard thing to do for XBTs).
- The best test from each QC group (defined as being the closest to 100% TPR and 0% FPR) is selected. 
- Other tests are then added using the existing method or trying to maximise the gradient of the ROC curve.
- An issue was identified with the ICDC spike check, where the test can return QC flags for fewer levels than it should. This has been fixed.

### DIFF
--- a/analyse-results.py
+++ b/analyse-results.py
@@ -1,15 +1,56 @@
+import csv
 import json, pandas
 import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
 import numpy as np
 import sys
+from util import wod_stats
 from util import dbutils, main
 
+def read_qc_groups(filename='qctest_groups.csv'):
+    # Read a csv file containing information on QC test groups.
+    csvfile = open(filename) 
+    groupinfo = csv.reader(csvfile)
+
+    # Create empty lists for each type of group.
+    removeabove    = []
+    removebelow    = []
+    removelevels   = []
+    removeprofile  = []
+    requiredgroups = {}
+
+    # Fill out the lists.
+    for i, spec in enumerate(groupinfo):
+        if i == 0: continue # Miss the header line.
+        if spec[1] == 'Remove above reject':
+            removeabove.append(spec[2])
+        elif spec[1] == 'Remove below reject':
+            removebelow.append(spec[2])
+        elif spec[1] == 'Remove rejected levels':
+            removelevels.append(spec[2])
+        elif spec[1] == 'Remove profile':
+            removeprofile.append(spec[2])
+        elif spec[1] == 'At least one from group':
+            if spec[0] in requiredgroups:
+                requiredgroups[spec[0]].append(spec[2])
+            else:
+                requiredgroups[spec[0]] = [spec[2]]
+        elif spec[1] == 'Optional':
+            continue
+        else:
+            raise NameError('Rule not recognised: ', spec)
+
+    csvfile.close()
+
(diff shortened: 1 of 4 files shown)


BillMills:
Nice catch on ICDC_07!

I'm running this myself right now, but just to cross-check: what sort of final TPR / FPR are you getting at the end of this?

s-good:
I'm typically getting about 80-85% TPR and 20% FPR from using one from each of the QC groups. The EN track check seems to find another 10% TPR but is costly in FPR so hopefully once the issues with that are sorted it will be useful. I am going to have a look at the track check soon.

BillMills:
@s-good that last 15% TPR is where the devil lives, though; trying to get TPR up to 100% will blow up your 20% FPR. Running this on the same sample we were using in Oostende, I get the following tests reported in `roc.json`:

```
  "tests": [
    "CoTeDe_GTSPP_spike_check",
    "CoTeDe_Argo_density_inversion",
    "CSIRO_short_gradient",
    "ICDC_aqc_05_stuck_value",
    "EN_increasing_depth_check",
    "WOD_range_check",
    "EN_background_available_check",
    "CoTeDe_GTSPP_WOA_normbias",
    "ICDC_aqc_08_gradient_check",
    "EN_std_lev_bkg_and_buddy_check",
    "ICDC_aqc_09_local_climatology_check",
    "CoTeDe_Morello2014",
    "EN_track_check"
  ]
```

If I omit the track check and OR the rest together, I get a TPR of 94% and an FPR of 63%; if I do not omit the track check, I get a TPR of 99% and an FPR of 91% (let me know if I'm interpreting `roc.json` incorrectly - I'm surprised there are no `&` combinations - ?). This performance is similar to what I'm getting from a data-driven approach that doesn't pre-select categories of tests, visible for comparison's sake in #226.

I'll continue to hammer on the edge cases like the track check - I think that is the most serious problem at the moment.

s-good:
As pointed out in [ this pull request](https://github.com/IQuOD/AutoQC/pull/226#issuecomment-399702450) we should discard tests that have poor discriminating performance. The change adds in a check that the ratio of TPR to FPR is greater than or equal to a limit (which defaults to 2). This is used in two ways:

1. A warning is printed to screen if a test has TPR/FPR below the limit but is the best out a group where at least one test has to be used.

2. After a QC test from each group is selected, any test with poor discriminating power is discarded.

I'm not clear in the first case whether we should discard a test if the TPR/FPR is low; however, if we do this we could be left without a QC test for a particular error mode so I think that we may have to use it.

BillMills:
> I'm not clear in the first case whether we should discard a test if the TPR/FPR is low

I would discard the QC test, or potentially use this in a future graduated QC approach - a profile getting flagged by such a test could be marked as 'probably bad' instead of 'definitely bad'. It's true we'll lower our true positive rate by throwing these tests away, but (cf the track check) I think we're scooping up a lot of profiles essentially randomly in these cases; the true positives we lose may well be bad for reasons actually unrelated to the QC test in question, and be candidates for future test development.

Anyway - that's all future work stuff. Let's merge here and see how we do on our upcoming data run.

PR #226, opened 2018-05-17
draft of data-driven algo to catch all bad profiles

BillMills:
Here is my algorithm for choosing combinations of QC tests that capture all bad profiles with a minimum FPR, which I mention in #225; this is optimized for aggressively pursuing 100% TPR, without relying on pre-defined classes of tests. I'm posting this mostly for comparison and record's sake - happy to go with @s-good's method, though as I describe below I think both suffer from the same underlying problem.

Running on our QuOTA sample constructed per our Oostende results, this produces a test combination list of:

```
  "tests": [
    "EN_constant_value_check",
    "Argo_global_range_check",
    "EN_range_check",
    "ICDC_aqc_06_n_temperature_extrema",
    "CSIRO_wire_break",
    "CoTeDe_spike",
    "ICDC_aqc_07_spike_check",
    "Argo_spike_test",
    "CSIRO_depth",
    "EN_increasing_depth_check",
    "ICDC_aqc_09_local_climatology_check",
    "EN_track_check",
    "CoTeDe_anomaly_detection",
    "CoTeDe_Argo_density_inversion",
    "CoTeDe_Morello2014",
    "CoTeDe_GTSPP_WOA_normbias",
    "Argo_regional_range_test",
    "CoTeDe_location_at_sea_test&CSIRO_short_gradient",
    "CoTeDe_GTSPP_global_range"
  ]
```

resulting in a TPR of 99.7% and a FPR of 97.2% - slightly higher on both counts than @s-good's method, and way way too high of an FPR to be ready for prime-time.

As I mention in #225, the difference between these two methods is largely personal taste, and either is acceptable - but they're both going to suffer from the same underlying problem: there is a class of profiles that are flagged by exactly one QC test, which also has a very high FPR, so if we push our TPR to 100% as we discussed, we're stuck with a > 90% FPR. `EN_track`, `CoTeDe_anomaly_detection`, and `CSIRO_depth` (in that order) are the worst offenders of this kind in the above run.

### DIFF
--- a/catchall.py
+++ b/catchall.py
@@ -1,0 +1,128 @@
+# algorithm:
+# 1. remove from consideration any bad profile not flagged by any test; put these aside for new qc test design
+# 2. accept all individual qc tests with 0% fpr; remove these from consideration, along with all profiles they flag
+# 3. form list of ntuple AND combos, add their decisions to consideration
+# 4. identify profiles flagged by exactly one combination. Accept that combination, drop all profiles marked by this combination, and drop this combination from further consideration
+# 5. drop the remaining combination with the highest false positive rate (note at this step all remaining profiles are flagged by at least two combination, so this will not raise the false negative rate).
+# 6. go back to 4; loop until the list of accepted combinations flags all bad profiles not dropped in step 1.
+
+import util.main as main
+import util.dbutils as dbutils
+import itertools, sys, json
+from operator import itemgetter
+
+def ntuples(names, n=2):
+    '''
+    given a list of names of tests, form every ntuple up to and including n combinations from the list
+    return as a list of tuples.
+    '''
+
+    combos = []
+    for i in range(2,n+1):
+        combos += itertools.combinations(names, i)
+
+    return combos
+
+def amend(combo, df):
+    '''
+    add a column to df describing the results of combo
+    column title will be the combo elements joined with '&'
+    '''
+
+    decision = df[combo[0]]
+    for test in combo[1:]:
+        decision = decision & df[test]
+
+    name = '&'.join(combo)
+    return df.assign(xx=decision).rename(index=str, columns={'xx': name})
+
+# get tests names and dataframe
+testNames = main.importQC('qctests')
+df = dbutils.db_to_df(sys.argv[1], n_to_extract=sys.argv[2])
+# declare some downstream constructs
+accepted = []
+unflagged = []
+fprs = []
+bad = df.loc[df['Truth']]
+bad.reset_index(inplace=True, drop=True)
+
+# mark chosen profiles as part of the training set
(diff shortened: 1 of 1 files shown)


s-good:
This looks like a really interesting method! If I understand correctly, it is finding the minimum set of QC tests needed to reject all the bad profiles that can be found using the QC tests we have coded up. It is good to have more than one method to find our set of checks as if they give consistent answers it gives us confidence in our results. I guess one thing that could be tried is weighing up the advantage of accepting a test that is uniquely finding some bad profiles against the disadvantage of the number of good profiles it incorrectly flags. This might give us a combination with a lower FPR but hopefully still a high TPR.

s-good:
This looks like it is working well! Shall I merge it or do you plan to add any more changes?

castelao:
Hi, what happened with CoTeDe_fuzzy_logic? From one of Simons' slides in Oostend, it was the best result on that point, correct?

BillMills:
@castelao are you referring to the result in the top post above? That was before imposing the TPR/FPR cutoff threshold I suggest above; after imposing that cutoff, `CoTeDe_fuzzy_logic` comes roaring back:

```
{
  "tests": [
    "ICDC_aqc_05_stuck_value",
    "Argo_global_range_check",
    "EN_range_check",
    "CoTeDe_spike",
    "ICDC_aqc_07_spike_check",
    "Argo_spike_test",
    "EN_background_available_check",
    "ICDC_aqc_08_gradient_check",
    "AOML_climatology_test",
    "CoTeDe_fuzzy_logic",
    "CSIRO_short_gradient",
    "Argo_regional_range_test",
    "CoTeDe_rate_of_change",
    "CSIRO_long_gradient",
    "EN_std_lev_bkg_and_buddy_check",
    "CoTeDe_GTSPP_WOA_normbias",
    "ICDC_aqc_09_local_climatology_check_tune5",
    "EN_spike_and_step_suspect",
    "WOD_gradient_check",
    "AOML_spike_tune20",
    "CoTeDe_location_at_sea_test",
    "CoTeDe_tukey53H_norm",
    "EN_track_check",
    "WOD_range_check",
    "AOML_constant"
  ]
}
```

What likely happened in the original was that `Morello2014`, which is also a fuzzy logic test if I understood correctly, got chosen, and `fuzzy_logic` didn't add any additional correct flags on top of this. However, `Morello2014` has a very high false positive rate compared to `fuzzy_logic`, so the latter comes out on top again when a discrimination power metric is applied.

castelao:
Thanks for the explanation @BillMills .

`Morello2014` is not fully fuzzy logic, but it doesn't matter if it is giving better results. I'll review the coefficients that I used in my fuzzy_logic.

Good progress!

PR #227, opened 2018-05-29
preliminary improvements to en-track

BillMills:
I've been fiddling with `EN_track` since Oostende to try and address issues discussed in #225 and #226. While this PR doesn't quite get us where I'd like to be, it makes some necessary improvements:

 - Per @BoyerWOD's comments, cruises are identified including the 2-character country code, and profiles reporting aircraft platforms are dropped. Note that profiles reporting country code 99 (unidentified / error) are not considered at this time.
 - Previous implementation of this test checked the suitability of the first profile passed to the test for assessment by EN_track, but didn't impose the same conditions on the collection of other profiles found to share a track with the first; this was an oversight, corrected here.
 - Speed thresholds are doubled, to 30 m/s for ships and 4 m/s for buoys. This was to accommodate instances of profiles getting flagged that just barely failed the test; see DE901006 for an example of a perfectly reasonable looking cruise that was just a little too fast.

These changes take us on our QuOTA sample from a TPR / FPR of 61.8% / 74.7% previously to 17.6% / 16.3%; most of the reduction in gross flags is currently due to not considering profiles with country code == 99. TPR > FPR is at least superficially auspicious, but it remains TBD how this affects downstream test combination algorithms, and if and how we can do even better.

More work forthcoming, but this seems like at least small progress to begin with.

### DIFF
--- a/build-db.py
+++ b/build-db.py
@@ -27,7 +27,9 @@ if len(sys.argv) == 3:
                 time real,
                 lat real,
                 long real,
-                cruise text,
+                country text,
+                cruise integer,
+                ocruise text,
                 probe integer,
                 training integer,
                 """
@@ -132,8 +134,11 @@ if len(sys.argv) == 3:
         truth = encodeTruth(profile)
         p['truth'] = main.pack_array(truth)
 
-        # extract country code to prefix cruise
+        # extract country code
         country = profile.primary_header['Country code']
+
+        # originator cruise
+        orig_cruise = profile.originator_cruise()
 
         # keep tabs on how many good and how many bad profiles have been added to db
         # nowire == index of first wire break level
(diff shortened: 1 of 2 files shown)


BillMills:
- Cruise and country separate in the database: sure, will do.
 - Probe type is how we decide if this measurement is a buoy or not; see the `isBuoy` function, which is pulling it out of a row of a data frame (which is why a quick grep won't turn up any other reference to probes - but that list of integers in `isBuoy` are all probe types we assume are buoyed).
 - Parameterizing tests is not something well supported at the moment, since the main routine just runs each test once with whatever parameterization is defaulted into it; but, I think this would be a great feature to add in future. I'd do this by returning test-specific flag parameterizations for QC results rather than just true / false, but this would require quite a bit of reengineering. Since the looser parameterization doesn't ultimately produce much better results in terms of resolving power, let's leave it for future work in the interests of moving on to our production run this month.
 - Originator cruise ID: so just to make sure I understand correctly, you want to consider a cruise to be a collection of profiles that share country code, cruise ID, and originator cruise ID? Sure, can do.

s-good:
Thanks for making the changes. I think this will be as good as we can make it. 

With the originator cruise ID, the main thing is to make sure it is not 0. As I understand it, the originator cruise gets converted into a WOD cruise number. Null/missing cruise originator cruise numbers like 0 can get translated into 'real' looking cruise IDs. This happens with a originator cruise number of 0, for example. The new code change stops this from happening, so I think this should be good for merging.

PR #228, opened 2018-06-30
AOML gradient test

BillMills:
the first of many AOML tests to be ported over from ftp://ftp.aoml.noaa.gov/phod/pub/bringas/XBT/AQC/AOML_AQC_2018/codes/.

quickie XBT performance:

```
                       NAME OF TEST   FAILS     TPR     FPR     TNR     FNR
                              Truth   10172  100.0%    0.0%  100.0%    0.0%
                      AOML_gradient    8233   56.9%    4.3%   95.7%   43.1%
```


BillMills:
@s-good right, of course - while a depth inversion is pretty suspicious, if we're going to flag it we should do so in a separate test, or at least a separate stanza, rather than conflating it with the gradient logic; see my latest corrections which should be in line with your suggestions.

PR #232, opened 2018-07-22
first pass at aoml climatology

BillMills:
work in progress - still need to write some unit tests for this, and tidy up some more, but the bulk of the implementation is visible here.

### DIFF
--- a/qctests/AOML_climatology_test.py
+++ b/qctests/AOML_climatology_test.py
@@ -20,6 +20,7 @@ def test(p, parameters):
 
     for i in range(p.n_levels()):
 
+        # find best interpolated temperature and standard deviation at this depth
         if not isData[i]: continue
 
         interpTemp = interp_helper.temperature_interpolation_process(p.longitude(), p.latitude(), p.z()[i], depthColumns1, latLonsList1, lonlatWithTempsList1, False, "climaInterpTemperature")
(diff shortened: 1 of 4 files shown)


BillMills:
@s-good this is essentially ready for review; I'm not entirely happy with the level of unit testing validating this qc test, but I decided that in the interest of integrating potential future work from the AOML project, it would be best to not be too aggressive in refactoring their code into more atomic functions that can be meaningfully unit tested.

Running over a batch of xbt profiles just to see what comes out, I got:

```
                       NAME OF TEST   FAILS     TPR     FPR     TNR     FNR
                              Truth   10000  100.0%    0.0%  100.0%    0.0%
                   AOML_climatology    8093   71.6%    9.3%   90.7%   28.4%
```

as of a032c91; I'm going to run this again on 047d3af later this week to make sure none of the superficial changes made there caused any regressions, but nevertheless please have a look and let me know any concerns.

BillMills:
@s-good thanks for checking this out - I think we're ok on both your points, but by all means let me know if I've missed something in the following resoning:

1.  You're right we always use the default of `woa13_00_025.nc`, but I believe this is an annual file. I pulled this from the original AOML project at ftp://ftp.aoml.noaa.gov/phod/pub/bringas/XBT/AQC/AOML_AQC_2018/data_center/, where it is referred to in their parameter file ftp://ftp.aoml.noaa.gov/phod/pub/bringas/XBT/AQC/AOML_AQC_2018/data_center/ with the key `Climatology_All_Year_AN_SD` which makes it sound like a full-year climatology, and the original authors further reference this in a function called `clima_custom_annual_file_path_and_name` in ftp://ftp.aoml.noaa.gov/phod/pub/bringas/XBT/AQC/AOML_AQC_2018/codes/qc_checks/clima_checker.py. All of which is inferential and not necessarily correct, but seems to suggest this file is intended for use throughout the year.

2.  My comment saying that `get_index_and_next` 'bracket' the point of interest is only roughly correct. This function returns the closest level in the array to the point of interest, and either one level higher or one level lower; if the point of interest is closest to level *n* but slightly higher than it, the function will return levels *n* and *n+1*; if the point is slightly lower then level *n*, the function returns *n-1* and *n*. All of which is a bracket, iff the list of levels is monotonically increasing and the point of interest is within the range of the list. Edge cases aside, the implementation is verbatim the same as the original, except for collapsing their `maximum_index` and `minimum_index` functions, which differ only by a minus sign, into a single function.

s-good:
Great - thanks for checking those things! I've merge the pull request.

PR #235, opened 2018-08-09
remove old datastore package from en_track unit testing

BillMills:
Thanks to @s-good for catching this!


s-good:
I ran the nose tests locally and all runs fine now apart from the two runtime warnings that appear in the travis checks, so I will merge.

PR #237, opened 2018-08-10
Modifications for QuOTA production run

s-good:
These two changes are:

1. Set up the build-db.py program to ignore originator_profile_type (as QuOTA does not have this set) and only read data for months 1, 2, 3 and 6 (the months fully manually quality controlled in QuOTA). This is selected by adding an optional configuration keyword to the end of the calling string.

2. A small modification to AutoQC.py to allow processing to be run in batches. This is helpful when there is a limit on how long a process is allowed to run as it makes it easy to break the processing into shorter batches.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -83,14 +83,22 @@ if len(sys.argv)>2:
   uids = main.dbinteract(query)
   
   # launch async processes
+  if len(sys.argv) > 4:
+    batchnumber = int(sys.argv[3])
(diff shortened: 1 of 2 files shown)


BillMills:
Great, thanks for this @s-good! I'd like to pull the config out of `build-db.py`; rather than baking in config for every run, we can pass that in as command line parameters. That said, I'll just merge this now and send by what I mean in a PR later this evening.

PR #239, opened 2018-08-14
Fix interpretation of True/False in command line

s-good:
We recently changed build-db.py to allow setting a logical from the command line. The syntax used looked like it should work:

`bool(sys.argv[3])`

However, it turns out that anything except an empty string gets turned into True by this.

This is an option for an alternative:

`ast.literal_eval(sys.argv[3])`

It uses the ast built in Python module. It works, but I don't know if this is the best solution.

### DIFF
--- a/build-db.py
+++ b/build-db.py
@@ -178,13 +178,12 @@ if len(sys.argv) == 3:
     
 elif len(sys.argv) == 5:
 
-    months = sys.argv[4].split(',')
-    months = [int(x) for x in months]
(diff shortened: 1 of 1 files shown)


BillMills:
Ah yes of course, my mistake - but in that case, we might as well also use it for the list of months; `ast.literal_eval("[1,2,3,6]")` instead of the split-parsing in the couple lines above. That way the command line arg can be entered like a normal python list.

s-good:
I've made that change and added an example of how to format the commands.

PR #241, opened 2018-08-31
dont ask for more points than possible from kdtree

BillMills:
@s-good in our recent data run, I see a grand total of zero flags raised by AOML climatology. Digging in a bit, it seems that `nearest_indices` in `util/AOMLinterpolation.py` is using scipy's kdtree to ask for the 100 closest lat/long points from a pre-determined lattice closest to the profile's position, but there's no protection to make sure there are actually at least 100 points in the lattice to return. Ask for more neighbors than the lattice has, and kdtree just returns a bunch of infs (see example below). Also, why 100? It seems to only take the closest 1.

This PR forces a sensible return from kdtree, and only asks for the nearest neighbor.

Minimum example of bad behavior in kdtree:
```
import numpy as np
from scipy.spatial import cKDTree

ll = [[0,0],[0,0.5],[0,1],[0.5,0],[0.5,0.5],[0.5,1.0],[1.0,0],[1.0,0.5],[1.0,1.0]]
tree = cKDTree(ll)

print tree.query([0.5,0.5], 9)
print tree.query([0.5,0.5], 10)
```

### DIFF
--- a/util/AOMLinterpolation.py
+++ b/util/AOMLinterpolation.py
@@ -97,7 +97,7 @@ def nearest_indices(y, x, latLonList, amountNearNeighbors):
   """
 
   tree = spatial.cKDTree(latLonList)
-  return tree.query([y,x], amountNearNeighbors)
+  return tree.query([y,x], min(amountNearNeighbors, len(latLonList)) )
 
 def temperature_interpolation_process(x, y, depth, depthColumns, llList,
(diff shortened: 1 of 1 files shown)


BillMills:
I have our full data run database with `AOML_climatology` re-calculated based on these changes; if the changes look good, I'll make it available through S3, let me know.

PR #243, opened 2018-10-07
Best QC set update

s-good:
This updates the code to generate best sets of QC tests:

1) Bug fix to the code to extract profiles into a data frame; before this was not deleting profiles that had been entirely rejected e.g. owing to having an impossible date or location so they were still being used for training and validation. The truth QC flags are now interpreted better: only flag 3 or 4 is deemed as a reject as before anything >= 3 was a reject which could include values such as 5 which means that the data value was changed and so the data value should be good.

2) Set up catchall.py to filter the profiles according to pre-chosen sets of tests e.g. XBT data <3.6 m depth and anything below an XBT wire break are now not considered.

3) analyse-results.py has a modification to the way the best sets are determined. It now has two default operating modes - one which gives a compromise set of QC tests (balancing maximising TPR with minimising FPR) and a conservative set (emphasising minimising FPR over maximising TPR) (the other possibility - emphasising maximising TPR - is done using catchall.py).


util/dbutils.py
```diff
@@ -28,7 +28,7 @@ def parse(results):
 def summarize_truth(levels):
     'given an array of originator qc decisions, return true iff any of the levels are flagged, ignoring t=99 levels'
 
-    return numpy.sum( [not levels.mask[i] and x>=3 and x<99 for i, x in enumerate(levels)]) >= 1
+    return numpy.sum( [not levels.mask[i] and (x==3 or x==4) for i, x in enumerate(levels)]) >= 1
```
BillMills: there are some flag sets that use higher numbers for problems (ie 7 is a 'known error' in GEOSECS data and PMEL TAO / PIRATA indicates 5 is a sensor failure) - but I'm not sure how relevant these are for our purposes. I'll take it on your recommendation to go with this for now, but we may have to revisit in future depending on where our training data is coming from.

catchall.py
```diff
@@ -41,10 +42,15 @@ def amend(combo, df):
...
-# get tests names and dataframe
-testNames = main.importQC('qctests')
+# Read QC test specifications if required.
+groupdefinition = ar.read_qc_groups()
 
-df = dbutils.db_to_df(sys.argv[1], n_to_extract=sys.argv[2])
+# Read data from database into a pandas data frame.
+df = dbutils.db_to_df(sys.argv[1],
+                      filter_on_wire_break_test = False,
+                      filter_on_tests = groupdefinition,
```
BillMills: this is great, thanks for this!

PR #244, opened 2018-12-18
make sure drop list isn't reset every action in db_to_df

BillMills:
@s-good sorry I've been a bit quiet for a while, I was chasing around why I was getting a bunch of surface XBTs bloating my false negatives when they should have just been masked out - turns out that for multi-action filters (like the default one), the list of profiles dropped for having no valid levels after filtration was being reset every action - so only the last action actually causes profiles to get dropped in `db_to_df`.

Also, I've turned the drop list into a set, just to make sure we don't try and drop the same thing twice (especially dangerous when dropping by table row ID).

### DIFF
--- a/util/dbutils.py
+++ b/util/dbutils.py
@@ -126,6 +126,7 @@ def db_to_df(table,
                     qc = qc[:nlevels[i]]
                 df.iat[i, j] = main.pack_array(qc)
 
+    todrop = set()
     for action in filter_on_tests:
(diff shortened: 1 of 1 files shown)


s-good:
Thanks for spotting this. The code changes look good.

 It looks like it will have affected the results I generated before so I will need rerun them. I've not had any time at all to work on this recently so it won't be too big a job to re-generate them.

PR #245, opened 2019-03-04
testing for bottom hits

BillMills:
~work in progress - not ready for merge~

Here's a sketch of a QC test that flags any level deeper than the ocean floor at the reported lat/long. Of the roughly 300 bad quota profiles we are currently failing to flag in our test sample (ie false negatives), about 50 are flagged by this test. This will likely be a part of the 'pre-cleaning' step where we discount obviously bad levels, like what we currently do for wire breaks.

While very successful at eliminating a large class of our remaining false negatives, this implementation relies on the google maps API to determine the depth of the sea floor, which is cost prohibitive ($5 USD / 1000 requests); a mergable implementation needs to be able to make this determination for free. @BoyerWOD, I see that NOAA provides bathymetry data at ex https://maps.ngdc.noaa.gov/viewers/bathymetry/; is there a machine-readable API version of this data?

### DIFF
--- a/qctests/IQUOD_bottom.py
+++ b/qctests/IQUOD_bottom.py
@@ -1,14 +1,24 @@
 import numpy
 import util.main as main
+from netCDF4 import Dataset
 
 def test(p, parameters):
+
+    # register bathymetry data in parameters object on first use
+    if 'bathymetry' not in parameters:
+        print 'extracting bathymetry'
+        nc = Dataset('data/ETOPO1_Bed_g_gmt4.grd')
+        parameters['bathymetry'] = {}
+        parameters['bathymetry']['longitude'] = nc.variables['x'][:]
+        parameters['bathymetry']['latitude'] = nc.variables['y'][:]
+        parameters['bathymetry']['depth'] = nc.variables['z'][:]
 
     d = p.z()
     lat = p.latitude()
     lon = p.longitude()
 
     # get ocean depth at this point
-    floor = -1.0*main.find_depth(lat, lon)
+    floor = -1.0*main.find_depth(lat, lon, parameters['bathymetry'])
 
     # initialize qc as a bunch of falses;
     qc = numpy.zeros(p.n_levels(), dtype=bool)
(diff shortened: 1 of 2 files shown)


BoyerWOD:
that bathymetry data from the noted link is the raw cruise data.  A better source may be https://www.ngdc.noaa.gov/mgg/global/.  This is gridded means (bathymetry and altimetry) over 1 arc minute (1/60th of a degree) boxes globally.  Not always accurate for a single geographic point especially in coastal areas, but easier to handle than raw data.

BillMills:
Thanks, @BoyerWOD - I've updated this implementation to use NOAA's bathymetry, which produces comparable results to google's, so that solves that piece. 

But now I'm not too happy with the speed of this implementation, probably due to the fact that it re-loads that netcdf file for every profile; I'll find some time to tidy that up, and we should be good to go here.

BillMills:
78dac1a speeds this test up by reading the bathymetry data once, and registering it on the `parameters` dictionary that gets passed between tests. With that, I'm finally happy with this implementation; @s-good please let me know what you think.

castelao:
It is not the fastest way to do that, but you already have [OceansDB](https://github.com/castelao/oceansdb) installed in your image. It interpolates the bathymetry, and WOA for any point. It can take the nearest point or interpolate for the exact (lat,lon) that you requested.

CoTeDe uses that for the at_sea check and Climatology comparison with WOA and CARS.

```
>>> import oceansdb
>>> db = oceansdb.ETOPO()
>>> h = db['topography'].extract(lat=17.5, lon=0)
```

BillMills:
Awesome, thanks @castelao! I would rather do it your way to avoid importing bathy data twice; I’ll reimplement using oceansdb and update this PR as appropriate.

BillMills:
a515584 reimplements using oceansdb bathymetry per @castelao's suggestion; results are qualitatively similar to the Google and NOAA bathys.

s-good:
The test looks good. The only thing I'm not sure about is whether the whole profile should get rejected as exceeding the bathymetry might point to a location error. However, I guess there must be circumstances where there is a depth error and the rest of the profile is OK, so it's probably wiser not to do that.

PR #247, opened 2019-05-23
New ICDC check

s-good:
This change implements a new ICDC climatology check.

The existing ICDC 09 check has also been changed to use the parameter store, and QC test requirements and groups updated.


BillMills:
The code changes look good to me; of course the tests aren't passing right now only because we need to include the new data file in our testing environment (but they pass just fine when I download those and re-run the tests locally). I like that you've included the code for translating from the original text-based data file to a netcdf file; I'm working on rebuilding the testing environment to include the new data right now, but I'm planning on including only the downstream netcdf file there since it is 1 GB instead of 14. Please let me know any objections.

Once that environment is rebuilt and tests are passing on travis, I think we can merge here without further code changes; I'm working on that actively right now and hope to have it done this week.

PR #248, opened 2019-05-28
python2->3 port

BillMills:
Work in progress, please do not merge yet - just posting to show some progress. Still to complete:

 - [x] Get tester sample running with no new complaints _(complete at 449113c)_ 
 - [x] Pass unit tests _(complete at c49ef92)_
 - [x] Run on 5k profiles in old and new versions and demonstrate no unjustified change in results _(complete at eaf86f9, but see comments)_
 - [x] Update supporting infrastructure (travis, install scripts, dockerfile). _(complete at 18a6bd6 )_


BillMills:
Running on `data/quota_subset.dat`, I see some differences in results for:

 - CoTeDe_GTSPP_WOA_
 - CoTeDe_Morello2014 
 - CoTeDe_fuzzy_logic
 - IQUOD_bottom (which relies on `oceansdb`)

after upgrading from:

 - `cotede`  0.18.0 -> 0.19.2
 - `oceansdb` 0.6.0 -> 0.8.6

as part of our py2->py3 upgrade.

Downgrading to `oceansdb` 0.6.0 makes IQUOD_bottom match, so I'm confident that's just a change (and probably an improvement) in the underlying topographical data used by `oceansdb`. But, python 3 requires `cotede` 0.19.x, which in turn throws errors without `oceansdb` 0.8.6, so both upgrades are necessary.

None of which is necessarily a problem - I just generally prefer to migrate a codebase with absolutely no changes in results, and then later separately make decisions on upgrading packages, so we can distinguish between changes due to upgrades, versus changes due to regressions.

I'm going to go ahead with larger scale tests, but @castelao please let us know what you think - should we be not surprised by this change in result, or does this seem suspicious to you?

BillMills:
After running on 5k profiles with the code at b848428 and the package loadout below, I get the same results for all qc tests except for (`<` are py2 results, `>` are py3):

```
22c22
<           CoTeDe_GTSPP_WOA_normbias    1779   68.2%   10.5%   89.5%   31.8%
---
>           CoTeDe_GTSPP_WOA_normbias    1510   59.0%    8.0%   92.0%   41.0%
27,29c27,29
<                  CoTeDe_Morello2014    3118   58.1%   65.6%   34.4%   41.9%
<                 CoTeDe_WOA_normbias     788   35.5%    0.6%   99.4%   64.5%
<            CoTeDe_anomaly_detection    2700   68.0%   43.2%   56.8%   32.0%
---
>                  CoTeDe_Morello2014    3349   62.9%   70.1%   29.9%   37.1%
>                 CoTeDe_WOA_normbias     685   31.0%    0.4%   99.6%   69.0%
>            CoTeDe_anomaly_detection    2709   68.1%   43.5%   56.5%   31.9%
31c31
<                  CoTeDe_fuzzy_logic    1623   65.7%    6.9%   93.1%   34.3%
---
>                  CoTeDe_fuzzy_logic    1681   65.8%    8.9%   91.1%   34.2%
33c33
<         CoTeDe_location_at_sea_test      31    1.1%    0.2%   99.8%   98.9%
---
>         CoTeDe_location_at_sea_test      28    0.9%    0.3%   99.7%   99.1%
46c46
<                      EN_track_check     146    5.7%    0.7%   99.3%   94.3%
---
>                      EN_track_check     148    5.8%    0.8%   99.2%   94.2%
55c55
<                        IQUOD_bottom     566   17.2%    6.8%   93.2%   82.8%
---
>                        IQUOD_bottom     588   17.9%    7.0%   93.0%   82.1%
``` 

 - **EN_track**: I'm confident this is actually an improvement; near the end of the track check, we mark all profiles on track as bad if more than half of them are; but 'half' is determined with py2's default integer division, and therefore comes out not as intended in some rare cases. py3's default floating point division matches intention and produces the small discrepancy above.
 - **IQUOD_bottom** is to be expected due to the change in `oceansdb` version and an update to the underlying topographical data.
 - **CoTeDe suite**: it's less clear to me precisely why these are different, but it likely corresponds to the non-optional package update mentioned in my comments above.

Current python package loadout:

```
root@e45e27c1b008:/# pip3 freeze
asn1crypto==0.24.0
cftime==1.0.3.4
Click==7.0
cotede==0.19.2
cryptography==2.1.4
cycler==0.10.0
decorator==4.4.0
filelock==3.0.12
future==0.17.1
gsw==3.0.3
idna==2.6
intel-openmp==2019.0
keyring==10.6.0
keyrings.alt==3.0
kiwisolver==1.1.0
matplotlib==3.1.0
mkl==2019.0
netCDF4==1.5.1.2
networkx==2.3
nose==1.3.7
numpy==1.11.3
oceansdb==0.8.6
pandas==0.21.1
pyArgo==0.0.4
pycrypto==2.6.1
pygobject==3.26.1
pyparsing==2.4.0
pyproj==2.1.3
python-dateutil==2.8.0
pytz==2019.1
pyxdg==0.25
scikit-fuzzy==0.4.1
scipy==0.18.1
seabird==0.11.0
SecretStorage==2.3.1
Shapely==1.6.4.post2
six==1.11.0
supportdata==0.1.2
wodpy==1.6.1
```

durack1:
@BillMills just out of curiosity, what is the cause of the order 3% differences between the py2 and py3 implementations? Are these different libraries being used, or is it literally the difference between default type changes in py2 -> py3?

BillMills:
@durack1 different libraries. Specifically, see my comments above on the underlying `CoTeDe` and `oceansdb` libraries; the ones we were using in py2 (0.18.0 and 0.6.0) don't work in py3, forcing us to upgrade. 

The discrepancy in CoTeDe results is likely due to this, but per my question to @castelao above I'd like to understand a bit more clearly on precisely what changed, if possible. It's easy to confirm yourself, for example, that looking up the sea floor depth using `oceansdb` 0.6.0 vs. 0.8.6 produces slightly different results for the same lat/long pair, but this is due to an update in the underlying data - so an improvement if anything. Are the other differences similarly attributable to updated underlying parameters, or are there implementation differences in CoTeDe? To be determined.

durack1:
@BillMills thanks for clarifying. I'm curious about the py2 -> py3 switch, and agree that as dependencies (and their underlying data) change such "improvements" will likely occur, but keeping a track of how and what causes the changes is a challenge.

BillMills:
@durack1 right - with python 2.7 eol coming in just a few months, I think a bunch of people are dealing with this now. As much as possible when migrating codebases, I like to demonstrate no change at all when changing underlying infrastructure (after all, there's no reason why you can't do the same thing in py3 as in py2, or in language x as in language y). But tightly coupled dependencies can make things more interesting; anyway, what are you working on? Happy to discuss, cc me there if that's useful.

BillMills:
If in our py3 environment I:

 - drop `oceansdb` back to 0.6.0
 - manually replace the source for `cotede/qctests/location_at_sea.py` with what it was in 0.18.0 (available https://github.com/castelao/CoTeDe/blob/1196aaccb60eb07cf0276786a6bf2d47958d24f2/cotede/qctests/location_at_sea.py) while leaving the rest of cotede at 0.19.2

I get the same result for `CoTeDe_location_at_sea_test` as I did in py2. Therefore, the discrepancy noted for this test above is either due to:
 - `oceansdb` update (which I'm still assuming is an improvement in the underlying dataset), or
 - code changes made to cotede for this specific test between 0.18.0 and 0.19.2, which I'm also inclined to trust as being equivalent or better.

Since this list consists only of likely-trustworthy sources and doesn't implicate any other parts of the AutoQC infrastructure, I'm satisfied that the changes to `CoTeDe_location_at_sea_test` are safe.

I'll try a similar method for the remaining discrepancies (`CoTeDe_GTSPP_WOA_normbias`, `CoTeDe_Morello2014`, `CoTeDe_WOA_normbias`, `CoTeDe_anomaly_detection` and `CoTeDe_fuzzy_logic`) in the coming days.

durack1:
@BillMills thanks, you're right with dependencies, python provides an incredible array of libraries to build on, but trying to maintain consistency with constantly changing versions is a challenge. I have been working on https://github.com/PCMDI/pcmdi_metrics amongst other things and have found a similar challenge maintaining consistent results as conda environments are updated and dependencies fix bugs

BillMills:
@durack1 yes that's a common problem; that's why tooling like a virtualenv, VM or docker container to capture your environment need to be a part of the software stack for scientific applications that demand reproducibility. Then you can even distribute that environment along with any papers or conference talks to support results presented, and pull in patches in a controlled fashion.

BillMills:
Similar to the above comments for CoTeDe's location at sea test, I can get the same results in our py3 environment as in our py2 environment for `CoTeDe_GTSPP_WOA_normbias`, `CoTeDe_Morello2014`, `CoTeDe_WOA_normbias`, `CoTeDe_anomaly_detection` and `CoTeDe_fuzzy_logic`, if:

- Drop oceansdb to 0.6.0 as above 
- Replace the following files in cotede 0.19.2 with their equivalents at cotede commit 1196aacc:
  - cotede/qctests/fuzzylogic.py
  - cotede/fuzzy/defuzz.py
  - cotede/fuzzy/fuzzy_core.py
  - cotede/fuzzy/membership_functions.py
  - cotede/qctests/woa_normbias.py
  - cotede/qctests/morello2014.py
  - cotede/anomaly_detection/anomaly_detection.py
  - cotede/qctests/anomaly_detection.py
  - cotede/qc.py, but just the block dealing with `if 'anomaly_detection' in  cfg:`
- typecast all calls to `array.keys()` to lists

The first point is the same assumed change in underlying parameterization; the last point is a trivial change in the return type of that array method between py2 and py3. The list of cotede files that need to be dropped back to their equivalent at 0.18.0 circumscribes where the relevant changes can be; per my comments above, there's a good chance that these changes reflect improvements in cotede between 0.18.0 and 0.19.2, but nonetheless bears some close examination to determine exactly what changed to be absolutely certain we understand why these results are changing before we accept them.

BillMills:
Per my amended comments above, I'm certain that `CoTeDe_location_at_sea_test` is only changing due to the change in `oceansdb` But, see my question in https://github.com/castelao/oceansdb/issues/13; there are some wild altitude differences between the two versions in the antarctic, which do affect our results (a tiny bit). This may be due to the presence / absence of ice, either in reality or in what's getting counted in oceansdb; I'd like to clearly understand which.

BillMills:
To get a bit more specific about what happened between `oceansdb` 0.6.0 and 0.8.6, see https://github.com/castelao/oceansdb/blob/master/oceansdb/datasource/etopo.json; the default resolution for topography data changed from 5min to 1min, and more importantly the 1min topography came from NOAA rather than oos.soest.hawaii.edu; there seem to be important differences between these topo files besides just resolution, per the above discussion.

Knowing that, I can get consistent py2/py3 results for `CoTeDe_location_at_sea_test` with all the latest cotede and oceansdb versions just by changing the default resolution back down to 5min, and therefore back to the original parameter file. But, this has no effect on the five remaining CoTeDe tests that need to be reconciled; therefore, there are implementation differences for these tests besides the change in oceansdb.

BillMills:
All 5 remaining CoTeDe discrepancies seem due to changes in the `interpolate` function in `/usr/local/lib/python3.6/dist-packages/oceansdb/woa.py`. Drop in the definition of this function and `subset` from `https://github.com/castelao/oceansdb/blob/7b59330be328daf01f4b9c682490444a8dd17b2e/oceansdb/woa.py`, and replace keys `mean` with `mn`; `standard_deviation` with `sd`; and `number_of_observations` with `dd` in `/usr/local/lib/python3.6/dist-packages/cotede/qctests/woa_normbias.py`, and the py3 stack matches the py2 results. @castelao would you expect different results from these two different implementations of woa's interpolate function?

That *should* do it - but I'm going to run a fresh set of 5k test profiles and make sure all the corrections described in this thread ameliorate discrepancies there, too.

BillMills:
Running on a fresh set of 5k profiles in the environment described by `install.sh` and captured in the container image `iquod/autoqc:ubuntu-18.04` produces the same behavior; at this point I'm confident this is a regression-free port to py3, up to the following summary of the long discussion above:

 - `EN_track`: a small difference resulted from the change from integer to floating point division when moving from py2 to py3, when determining whether half of profiles on track were flagged. The code always assumed the division was floating point, which was actually incorrect in py2; so, this change reflects an improvement in implementation for this test.
 - `IQUOD_bottom` and `CoTeDe_location_at_sea_test`: these change due to a change in underlying topography in the `oceansdb` package between 0.6.0 and 0.8.6. The default underlying data changed from the 5 degree to the 1 degree resolution file, and furthermore the 1 degree file appears to be a bedrock file while the 5 degree one (I assume) includes ice, since it reports much higher elevations at points in Antarctica.
 - `CoTeDe_GTSPP_WOA_normbias`, `CoTeDe_Morello2014`, `CoTeDe_WOA_normbias`, `CoTeDe_anomaly_detection` and `CoTeDe_fuzzy_logic`: all of these were affected by the change to the `interpolate` function in `oceansdb/woa.py`.

I'm inclined to believe that the updates to `oceansdb` and `cotede` reflect improvements in those packages, leaving us good to merge here. Comments please @castelao and @s-good; I think we can merge here if this all seems reasonable to you.

s-good:
Thanks @BillMills for doing this extensive set of changes to get python 3 working. I had a look through and it all looks reasonable to me, so I'm happy for the code changes to be merged when you are ready.

PR #256, opened 2019-11-25
use flags rather than positional arguments for all key scripts

BillMills:
The original purpose of this work was to make the output locations of all files user-settable, so they could be targeted at mounted directories for ease of use with containerized executions of AutoQC. The least-messy way to do this was to refactor all scripts to use flags rather than positional arguments; this way order of parameters no longer matters, and we don't have to infer what mode we're running a script in by the number of parameters passed.

Besides that, it also became evident while working on this that `analyze-results.py` fails if all the qc tests in a given required set are eliminated for being nondiscriminating or not flagging any bad profiles; here I allow `analyze-results` to continue without picking anything from the class of tests in question in this case.

@s-good please let me know what you think of this approach, particularly with `analyze-results`.

### DIFF
--- a/AutoQC.py
+++ b/AutoQC.py
@@ -24,7 +24,7 @@ def process_row(uid, logdir, table='iquod', targetdb='iquod.db'):
   sys.stderr = open(logdir + "/" + str(uid) + ".stderr", "w")
 
   # extract profile
-  profile = main.get_profile_from_db(uid)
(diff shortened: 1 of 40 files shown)


BillMills:
Latest commits resolve regressions (1000 profiles produce the exact same results before and after this PR), and get tests passing - going to go ahead and merge here. As always, please post comments after merge with any concerns.

PR #257, opened 2020-06-24
Updates used in production run for paper

s-good:
This pull request includes some changes found while running AutoQC to generate results for the paper. They are:

- Change default number of iterations and cost ratio settings in analyse-results.py.
- Fix two incorrect entries in qctest_groups.csv.
- Disregard QC results from CSIRO_wire_break.py if it rejected the whole profile as this is very likely to be an erroneous result.
- Add an setting to util/dbutils.py to turn off the parsing of QC test results to give a single result for each profile. This is useful if plotting profiles and we want to know the QC result for each level in the profile.
- Reset the data frame index in util/dbutils.py after deleting a column. If this is not done the indexing of the data frame does not work as expected.


BillMills:
These changes all make sense, thanks for this! I'd like to take this opportunity to capture the state of the code used in the production run as exactly as possible by stamping a release to match and giving it a DOI we can cite like in wodpy - but, I assume the production run was done some months ago, almost certainly before PR #256 was merged just three weeks ago. Is it a safe bet that the production run was done:

 - after the port from Py2 -> Py3
 - before #256 (so at commit e498c594ac7e33d998e500be47aebf817a0f74ba)
 - but including the changes posted here?

If you're not sure and still have your production environment available, a `git log` in the AutoQC directory there may be revealing.

Anyway like I said, this looks great, I'll merge right now - but please let me know if you can on just exactly what the code state was for the paper so we can capture for posterity.

PR #258, opened 2020-08-01
[wip] package updates

BillMills:
- [x] numpy and scipy bump to address @s-good's comments re: https://github.com/scikit-learn-contrib/hdbscan/issues/272 *done in 574c6da*
 - [ ] ~rev cotede to 0.21.1, along with any necessary dependencies.~ *recommend doing this in a separate PR*
    - [x] oceansdb 0.8.6 -> 0.8.11; done in d613352. Note this causes one of the CoTeDe WOA unit tests to fail, presumably because this rev appears to move from woa13 to woa18 5deg temperature record files. Expected value changed to match by fiat in the unit test in the same commit; @castelao please let us know if a change from 0.8.6 to 0.8.11 should _not_ cause a results change.


BillMills:
There appears to be some non-trivial differences between CoTeDe 0.19.2 and 0.21.1 that come out in the glue script https://github.com/IQuOD/AutoQC/blob/master/cotede_qc/cotede_test.py that allows AutoQC to ingest CoTeDe. Specifically, the config object `cfg` has a completely different structure, having gone from a dictionary in 0.19.2 to an `orderedDict` in 0.21.1 which appears to have different keys than the original.

@castelao, can you advise on the mapping between old and new config objects? It's likely just a difference of some key names, but it'll take me a while to tease out the changes myself.

@s-good I would go ahead and merge this PR now, while we assess how to get CoTeDe upgraded; there's no need to block the other package updates on this.

castelao:
HI @BillMills , yes, OceansDB update will change some results. You're correct, that reflects Tim's update on WOA.

About the CoTeDe config, the new standard should recognize your config as an old standard and fix it in the background. Could you point me to the error message that you get?

castelao:
The [fix_config()](https://github.com/castelao/CoTeDe/blob/dee83c8837ec1d917e9a157eaf3c3465a5a71333/cotede/utils/config.py#L121) should conform your config files into the new standard. It is more than just renaming. The new standard is more explicit, and includes a 'rev' to help to track and fix future changes.

`new_standard_config = cotede.utils.config.fix_config(your_dict)`

But you shouldn't require to do that yourself. I probably missed something.

BillMills:
@castelao I think the trouble is that AutoQC has some logic hard-coded into it that assumes the config returned by CoTeDe's `load_cfg` function is the old style, which is not true if we upgrade past 0.19.2. Specifically see https://github.com/IQuOD/AutoQC/blob/master/cotede_qc/cotede_test.py#L49-L68 - this starts with a call to `load_cfg` which in CoTeDe 0.22.0 returns the new-style config, but the AutoQC code then goes on to treat that config like old-style config, resulting in unit test failures like the one below, which is really just a bunch of keyErrors:

```
======================================================================
ERROR: Make sure CoTeDe Argo density inversion is working as expected.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/AutoQC/cotede_qc/cotede_test.py", line 60, in get_qc
    cfg = {var: {test: cfg[var][test]}}
KeyError: 'TEMP'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/AutoQC/cotede_qc/cotede_test.py", line 65, in get_qc
    cfg = {'main': {test: cfg['main'][test]}, var: {'dummy': None}}
KeyError: 'main'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/AutoQC/tests/CoTeDe_checks_validation.py", line 33, in test_CoTeDe_Argo_density_inversion
    qc = qctests.CoTeDe_Argo_density_inversion.test(padi, None)
  File "/AutoQC/qctests/CoTeDe_Argo_density_inversion.py", line 10, in test
    qc = get_qc(p, config, testname)
  File "/AutoQC/cotede_qc/cotede_test.py", line 67, in get_qc
    with open('cotede_qc/qc_cfg/' + config + '.json') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'cotede_qc/qc_cfg/argo.json'
```

So there's work to be done not on CoTeDe but on how AutoQC interprets CoTeDe's config object.

BillMills:
Replaced by #264 and recent CoTeDe work.

PR #260, opened 2020-08-12
Gross range check

s-good:
This check was suggested by @castelao as a pre-filter to remove gross outliers before the other QC checks are run. It is similar to other global range checks but has an extreme range so only obviously bad data are excluded.


BillMills:
Looks good to me!

PR #261, opened 2020-11-10
cotede updates + CI passing

BillMills:
This PR extends and replaces #259 by updating the CoTeDe tests and CI environment to pass after making the changes suggested by @castelao in that PR.

We should expect and we do see some changes to the QC decisions made by CoTeDe due to the update to CoTeDe and oceansdb. The commits added to this PR that build upon #259 change the expected outcome for CoTeDe unit tests so that they pass by fiat, barring an a priori expected result. However, some of these changes are concerning; to sanity check them independently from the test suite, I ran AutoQC on 1000 profiles before and after this change. The CoTeDe results are as follows:

Before this PR:
```
                       NAME OF TEST   FAILS     TPR     FPR     TNR     FNR
      CoTeDe_Argo_density_inversion      81    9.4%    7.2%   92.8%   90.6%
          CoTeDe_GTSPP_WOA_normbias     295   59.2%    8.2%   91.8%   40.8%
          CoTeDe_GTSPP_global_range      62   14.6%    0.2%   99.8%   85.4%
              CoTeDe_GTSPP_gradient      80   19.2%    0.0%  100.0%   80.8%
       CoTeDe_GTSPP_profile_envelop     109   26.1%    0.0%  100.0%   73.9%
           CoTeDe_GTSPP_spike_check      26    6.2%    0.0%  100.0%   93.8%
                 CoTeDe_Morello2014     679   64.3%   70.5%   29.5%   35.7%
                CoTeDe_WOA_normbias     125   29.3%    0.5%   99.5%   70.7%
           CoTeDe_anomaly_detection     542   69.5%   43.2%   56.8%   30.5%
             CoTeDe_digit_roll_over     100   23.7%    0.2%   99.8%   76.3%
                 CoTeDe_fuzzy_logic     338   68.8%    8.7%   91.3%   31.2%
                    CoTeDe_gradient      85   20.4%    0.0%  100.0%   79.6%
        CoTeDe_location_at_sea_test       5    0.7%    0.3%   99.7%   99.3%
              CoTeDe_rate_of_change     161   30.9%    5.5%   94.5%   69.1%
                       CoTeDe_spike      10    2.4%    0.0%  100.0%   97.6%
               CoTeDe_tukey53H_norm      26    5.3%    0.7%   99.3%   94.7%
```

After this PR:

```
                       NAME OF TEST   FAILS     TPR     FPR     TNR     FNR
      CoTeDe_Argo_density_inversion      32    3.8%    2.7%   97.3%   96.2%
[...]

PR #262, opened 2020-11-15
minmax

BillMills:
Here's an implementation of minmax, from our colleagues at Oceanscope. A few thoughts on the first draft:

 - The helper functions provided by the original authors are included verbatim. Some simplification might be possible since that implementation folded running over lists of profiles into the test logic, whereas AutoQC factors this out.
 - Unit tests are currently somewhat minimal and could do with some expansion. Currently they check that the temperature mins and maxes and the final QC results come out the same in this implementation as they did in the example provided by the authors for the first profile in their example dataset. That's a good start, but I'd like to see checks on edge cases and more pathological profiles as well, including exception handling around missing data.
 - QC runs and tests pass in the environment described by the container image `iquod/autoqc:1.0-rc3`, with the following modifications:
   - upgrade pandas -> 1.1.4
   - install xarray at 0.16.1
   - copy original author's parameter files into `/AutoQC/data`
   I haven't posted this environment anywhere yet as @s-good mentioned we may need to discuss proper attribution before we distribute those parameter files. Perhaps a DOI should be created for them if feasible?
 - Domain experts: I wonder if it's possible to re-write this to not rely on salinity data, so that we can successfully run this test on profiles that only have temperature, depth, lat and long available. The only thing the salinity data object seems to be getting used for is its depth profile; perhaps we could take that elsewhere and do away with salinity considerations.

Let me know any thoughts on the above or the implementation included, and we can correct from there.


BillMills:
Many thanks to Rachel Killick for the testing, comments and implementation in 713269a!

BillMills:
Tests are now passing for minmax in image `iquod/autoqc:1.0-rc4`; the fails reported by CI are all actually CoTeDe fails, which are already resolved in #261.

I also include here an outline of a `REFERENCES.md` where we can list DOIs and other citations; CoTeDe and Wodpy have obvious DOIs to cite and minmax's code will also have one shortly per our discussion with Jerome. @s-good let me know what you think would be best for citing everyone else (AOML, CSIRO, ICDC and others).

s-good:
The draft paper contains the full list of citations for all the QC checks. We could either link to the paper once it is published or take the information from the paper and put it into a table. 

Should we merge this pull request and open a new issue to sort this?

BillMills:
Hi @s-good - you're right, let's leave the references out for now and link the paper once it's published; better to have one list in one place then reproduce everywhere.

In that case, I'll go ahead and merge here; let me know anything I can do to help get this paper over the finish line!

PR #263, opened 2021-01-25
Adding plain tukey53H

castelao:
The normalized version was already included in AutoQC. This is the absolute value check.


castelao:
@s-good and @BillMills , this is a WIP. I'll ping you again once I think it is ready to go.

BillMills:
Awesome, thanks @castelao, looking forward to it!

castelao:
@BillMills @s-good , does it work like that?

s-good:
Could you add to the CoTeDe tests as a way of confirming it works? E.g. add a duplicate of the Tukey53H norm test (https://github.com/IQuOD/AutoQC/blob/master/tests/CoTeDe_checks_validation.py#L147) (the line running the gradient test in L153 shouldn't be there although it won't affect the results).

castelao:
@BillMills @s-good , I think it is ready to go now. I tested using your latest docker.

BillMills:
Thanks for this, @castelao - I think the tests are failing on something simple; here you're providing a dictionary for the config, where all other AutoQC CoTeDe usage used strings for configs. If we have a look at https://github.com/IQuOD/AutoQC/blob/master/cotede_qc/cotede_test.py#L43, AutoQC's CoTeDe adaptor knows what to do with strings, but needs an else catch to just relabel `config` as `cfg` if a dictionary is passed in. With that change, tests pass - can you throw that in please, then I think we'll be good to merge here.

BillMills:
LGTM, thanks @castelao - @s-good, I'm happy to merge here when you are.

s-good:
Looks good - thanks @castelao and @BillMills. I will merge the pull request.

PR #265, opened 2021-05-13
Tidying up

s-good:
This pull request does some tidying in preparation for our production run:

- Remove warning message in AOML_spike.py due to using numpy.median on a masked array (the mask is already checked for missing data prior to the use of numpy.median so the warning message is not useful). Using numpy.ma.median prevents the warning being printed. Numpy.ma.mean was also used in place of numpy.mean, but I don't think this was giving warnings.
- Try to make all tests return a QC result for all levels. Some QC tests such as the track check were just returning a single value, which is inconsistent with what most of the tests are doing.
- Fix an issue that was occurring for the EN_std_level_bkg_and_buddy_check, which tries to use previously cached results from other tests. The code wasn't checking whether the result was present properly, causing the check to fail.
- Prevent a warning from the minmax test due to the presence of NaNs in the arrays. I have changed the code to check that the temperature, min and max are all present before attempting to make a QC decision and defaulting to a pass result if not.

### DIFF
--- a/qctests/EN_background_check.py
+++ b/qctests/EN_background_check.py
@@ -21,9 +21,10 @@ def test(p, parameters):
     # run the QC.
     #query = 'SELECT en_background_check FROM {} WHERE uid = {};'.format(parameters["table"], p.uid())
     #qc_log = main.dbinteract(query)
-    #if qc_log[0] is not None:
+    #if len(qc_log) > 0:
     #    qc_log = main.unpack_row(qc_log[0])
-    #    return qc_log[0]
+    #    if qc_log[0] is not None:
+    #        return qc_log[0]
 
     return run_qc(p, parameters)
 
--- a/qctests/EN_background_check.py
+++ b/qctests/EN_background_check.py
@@ -9,6 +9,7 @@ import numpy as np
 import util.obs_utils as outils
 from netCDF4 import Dataset
 import util.main as main
+from util.dbutils import retrieve_existing_qc_result
 
 def test(p, parameters):
     """
@@ -19,12 +20,12 @@ def test(p, parameters):
 
     # Check if the QC of this profile was already done and if not
     # run the QC.
-    #query = 'SELECT en_background_check FROM {} WHERE uid = {};'.format(parameters["table"], p.uid())
-    #qc_log = main.dbinteract(query)
-    #if len(qc_log) > 0:
-    #    qc_log = main.unpack_row(qc_log[0])
-    #    if qc_log[0] is not None:
-    #        return qc_log[0]
+    qc_log = retrieve_existing_qc_result('en_background_check', 
+                                         p.uid(),
+                                         parameters['table'], 
+                                         parameters['db'])
+    if qc_log is not None:
+        return qc_log
 
     return run_qc(p, parameters)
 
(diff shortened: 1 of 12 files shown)


qctests/EN_background_check.py
```diff
@@ -21,10 +21,10 @@ def test(p, parameters):
...
     #query = 'SELECT en_background_check FROM {} WHERE uid = {};'.format(parameters["table"], p.uid())
     #qc_log = main.dbinteract(query)
-    #qc_log = main.unpack_row(qc_log[0])
-
-    #if qc_log[0] is not None:
-    #    return qc_log[0]
+    #if len(qc_log) > 0:
+    #    qc_log = main.unpack_row(qc_log[0])
+    #    if qc_log[0] is not None:
+    #        return qc_log[0]
```
BillMills, line 27: This likely isn't meant to be commented out - looks like that was done long ago (by me when porting from python 2->3, sorry) and forgotten about until now.

qctests/EN_constant_value_check.py
```diff
@@ -17,9 +17,10 @@ def test(p, parameters):
     # run the QC.
     query = 'SELECT en_constant_value_check FROM ' + parameters["table"] + ' WHERE uid = ' + str(p.uid()) + ';'
     qc_log = main.dbinteract(query, targetdb=parameters["db"])
-    qc_log = main.unpack_row(qc_log[0])
-    if qc_log[0] is not None:
-        return qc_log[0]
+    if len(qc_log) > 0:
+        qc_log = main.unpack_row(qc_log[0])
+        if qc_log[0] is not None:
+            return qc_log[0]
```
BillMills, line 23: This block, from defining the query to deciding if something was successfully fetched from the DB, is both arcane enough and repeated enough to warrant its own helper function in utils, with a unit test or two. Doesn't have to block this PR, but would be a bit tidier.

BillMills (changes requested):
Thanks for this, @s-good; I particularly like standardizing on per-level QC instead of jumping back and forth, that's much better.

See a couple of minor comments inline; I think the one in `EN_background_check` is the only one that really needs to change before we merge. My point about factorization in `constant_value` is a nice to have - should be easy to implement, but can live in a future PR.

util/dbutils.py
```diff
@@ -192,3 +192,34 @@ def db_to_df(table,
...
+    table - the database table for the profile.
+    db    - the name of the database file.
+    '''
+    
+    query = 'SELECT {} FROM {} WHERE uid = {};'.format(test, table, uid)
+    qc_log = main.dbinteract(query, targetdb=db)
+    try:
+        if len(qc_log) > 0:
+            qc_log = main.unpack_row(qc_log[0])
+            print('QC results extracted from database:')
```
BillMills, line 210: Looking great after today's updates, thanks @s-good - the only nit I'll pick (and we're really getting down to little stuff) is I'd remove the `print` statements from the successful path. These will result in thousands of lines in the logs that don't tell us much, and can drown out logs of actual problems. In future we should probably use different logging levels from the `logging` package, but that's beyond the scope of this PR.

The print statements in the error paths are good though - those are the things we want to stick out.

BillMills:
I love it - thanks for the amendments, I say merge it and start the production run!

BillMills (approved):

s-good:
Great! Thanks for the comments. I'll merge it now.

PR #266, opened 2021-05-15
Further tidying

s-good:
Two changes to remove warnings/errors during productions runs:

1. Check for zero depth change in AOML_gradient to avoid divide by zero. I wasn't sure what to do in this situation, but went with setting the QC check to ignore pairs of levels with zero depth change as there are other QC checks that are designed to deal with this kind of problem.
2. Remove dropping of tables that store data produced by QC checks such as EN_background. This causes problems if running in batches as it deletes results from previous batches that are needed by the EN_std_level_bkg_and_buddy_check.

During investigating 2 I noticed something about the util/main.dbinteract function, which is that it doesn't call the database commit function. According to the sqlite3 documentation, this is needed else changes made to the database are not saved (https://docs.python.org/3/library/sqlite3.html). This is used in build-db.py but never when running AutoQC.py. Yet, clearly the results are being saved to the database, so maybe I don't understand this properly. @BillMills what do you think?

### DIFF
--- a/qctests/AOML_gradient.py
+++ b/qctests/AOML_gradient.py
@@ -17,7 +17,9 @@ def test(p, parameters):
 
     for i in range(p.n_levels()-1):
         if isData[i] and isData[i+1]:
-            gradTest = (t[i+1]-t[i]) / (z[i+1]-z[i])
+            dz = z[i+1]-z[i]
+            if dz == 0.0: continue
+            gradTest = (t[i+1]-t[i]) / dz
             if t[i+1]-t[i] < 0:
                 if abs(gradTest) > 1.0:
                     qc[i] = True
--- a/qctests/EN_background_check.py
+++ b/qctests/EN_background_check.py
@@ -204,7 +204,6 @@ def readENBackgroundCheckAux():
 
 def loadParameters(parameterStore):
 
-    main.dbinteract("DROP TABLE IF EXISTS enbackground", targetdb=parameterStore["db"])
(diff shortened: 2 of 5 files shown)


BillMills (approved):
Hi @s-good, thanks for this - all looks good to me. Some comments:

 - I like the separation of concerns in AOML_gradient; if other tests check for constant values, let them. Many narrowly focused tests are easier to implement, maintain, and draw conclusions from than a few overly-comprehensive monoliths.
 - I also agree with not dropping tables; this was a convenience for situations where we wanted to run build-db once and AutoQC multiple times, but this clashes with the batch mode implemented in AutoQC, as you pointed out. We should make sure we capture somewhere, either in the paper, the README, or probably both, what is expected when running the full pipeline, roughly:
     - raw data prep
     - running build-db
     - running AutoQC (here's where we note AutoQC should be run on fresh output of build-db, for example).
     - running postprocessing
  - Regarding SQL commits, thanks for pointing this out, that's an interesting point about database management I hadn't looked into carefully before. `main.dbinteract` works without explicitly asking for database commits thanks to this line right here: https://github.com/IQuOD/AutoQC/blob/master/util/main.py#L177, specifically the `isolation_level=None` option. This puts our database into autocommit mode, which does exactly what it sounds like, as noted here: https://docs.python.org/3.7/library/sqlite3.html#controlling-transactions.

I don't think any changes are required here - looks great, merge away!

s-good:
Great, thanks for the explanation about the dabase writing!

PR #267, opened 2021-05-18
Fix issues handling times

s-good:
Two fixes to prevent errors if there are missing values in the date information provided with a profile:

- Remove testing for year < 1900 in cotede_qc.py. Looking back into the code history, this was included because datetime.datetime.strftime could not handle dates before 1900, but as stated [here](https://docs.python.org/3/library/datetime.html#technical-detail) it has been able to do this since version 3.3. Therefore, we can safely remove these lines of code since AutoQC is using Python 3.6.
- In EN_std_level... check that year, month and date are not None before comparing them to integers.


BillMills (approved):
Looks great, thanks!

PR #268, opened 2021-05-22
adding zenodo metadata

BillMills:
Per @castelao's excellent suggestion, to capture author orcids ahead of release. 

@s-good that looks like your ORCID, but please double check; @castelao I lifted your entry right out of your example, lmk if that's all current and accurate.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,23 @@
+{
+  "title": "AutoQC: Automatic Quality Control analysis for the International Quality Controlled Ocean Database",
+  "creators": [
+    {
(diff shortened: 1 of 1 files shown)


castelao (approved):
If you want, you could also add "related_identifiers" and include wodpy's DOI as "references", as shown at https://github.com/castelao/inception.

s-good:
Hi, yes that's my ORCID.

castelao (changes requested):
Thanks for including CoTeDe. It would be best to use instead the DOI: 10.21105/joss.02063

PR #271, opened 2021-06-04
db_to_df batchtesting

BillMills:
Extends #270 with a unit test to make sure results don't change based on batch size in db_to_df.


s-good:
The new test looks good. Shall we merge it?

BillMills:
All of the above merged in #270 and #275.

PR #272, opened 2021-06-17
Upgrade to oceansdb

s-good:
Upgrade to oceansdb to solve an issue that was encountered with a small number of profiles.

Does the version number need changing anywhere else or is install.sh the only place it is set?


BillMills:
Thanks, @s-good - merged as part of #274, which also updates the containerized testing environment, as well as un-breaks travis after their recent server migration. CI should work as normal again, please let me know if not.

PR #273, opened 2021-06-17
package upgrade in testing environment

BillMills:
Expands on #272 to also upgrade oceansdb in our containerized environment, which is what the travis tests run in.


BillMills:
[It appears that travis has migrated from a .org to a .com](https://docs.travis-ci.com/user/migrate/open-source-repository-migration); I think I've got it re-wired-up correctly, but might need to make a new PR to kick off the machinery. I'm going to try closing and reopening this one to see if that's got it.

PR #275, opened 2021-07-05
test db_to_df batching

BillMills:
Exact same change as in #271, just cleaning up the already-merged commits.


BillMills:
(turns out Travis doesn't like it when you merge a PR while its trying to pull it, but as expected from tests passing in #271, the merged build passes despite the complaint above: https://app.travis-ci.com/github/IQuOD/AutoQC/builds/231839834)

PR #276, opened 2021-07-22
corrections to db management in catchall

BillMills:
Per the discussion at https://github.com/IQuOD/AutoQC/pull/270/files#r644709834, looks like maybe this correction didn't make it in and we missed it at the time.


s-good (approved):
Thanks for spotting and fixing this. It all looks good to me.

PR #277, opened 2022-01-07
Update to QC test group list

s-good:
Updates to the QC test group list to:

- Add the CoTeDe Tukey53H test.
- Change the rule for negative depths to 'remove above reject'. This was just to avoid problems when using the 'XBTbelow' functionality in db_to_df, as otherwise a whole profile could get removed.


BillMills:
Thanks very much for this, @s-good - I'll tag and we can reference this in our forthcoming paper, as discussed.

PR #279, opened 2022-07-14
fixes #278

### LINKED ISSUE #278: EN stability check minor bug
In the EN stability check, the density difference between two levels is supposed to be calculated at a single pressure level (i.e. difference in potential density) as shown in the equation at the bottom of https://www.metoffice.gov.uk/hadobs/en3/OQCpaper.pdf page 8. Currently the difference calculated is between the density at their respective levels e.g. https://github.com/IQuOD/AutoQC/blob/master/qctests/EN_stability_check.py#L53.

A fix was tested on QuOTA data but the difference in results was found to be negligible (2 more profiles rejected out of 47022 profiles processed), therefore this is a low priority fix.

BillMills:


BillMills:
Thanks for catching this, @s-good - please let me know if there's more to it than this simple correction.

s-good (approved):
Thanks for making the changes - they are exactly what is needed as far as I can see.

PR #280, opened 2023-01-07
[wip] iquod flagging script

BillMills:
This PR will contain a script to add a column to AutoQC's sql database that indicate per-level IQuOD QC flags for the profiles analyzed. This script is meant to run as a standalone post-processing step, similar to `summarize_results.py`.

### DIFF
--- a/iquod_flags.py
+++ b/iquod_flags.py
@@ -1,0 +1,119 @@
+import numpy, pandas, sqlite3, math
+import util.main as main
+import util.dbutils as dbutils
+
+def parse(results):
+    'lifted from dbutils without the summary'
+
+    return results.apply(dbutils.unpack_qc)
+
+def db_to_df_simplified(table,
+             n_to_extract=numpy.iinfo(numpy.int32).max,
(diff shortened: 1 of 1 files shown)


BillMills:
To be resolved before merge:

 - @s-good, I assume "CS XBT surf. temp." referred to in the paper corresponds to the QC test`CSIRO_depth`; let me know if this is wrong.
 - Speaking of `CSIRO_depth`, including this saddles many profiles in `quota_subset` with flag 3 for every level, since we're propagating XBT flags downwards. Is this really what we want, or in this case is just the surface measurement bad?
 - Todo: still need to actually write the result back to the database
 - Todo: need a better validation sample to make sure I'm getting what we expect. @BoyerWOD if you have or can generate some wodascii with correct IQuOD flags we can test against, that would be helpful.

s-good:
Sorry for the delay in replying to this.
* Yes, CS XBT surf. temp. is CSIRO_depth.
* CSIRO_depth should not be propogated downwars. The same is true for ICDC_aqc_01_level_order.

Issue #281, opened 2024-01-26
EN_stability_check use of pressures


s-good:
The EN_stability_check code requires pressure data, which it accesses from the profile data using the p() function, which in turn gets the data from depth-dependent primary variable code 25 in the WOD ASCII data. This is not always available, though, and that means the check cannot run. If there are no pressure data available, the code should attempt to convert them from the depth data. This will allow the test to run on more profiles than it can at present.

PR #282, opened 2024-02-09
en_stability fallback

katieannemills:
@s-good - here's about the simplest possible attempt at #281, let me know if we're looking for something more sophisticated.


s-good:
Thanks for working on this. Another alternative is to use the function p_from_z() from gsw (https://teos-10.github.io/GSW-Python/conversions.html) which we use already in a number of places in AutoQC. I expect there is little difference between them, though. @BoyerWOD, do you have a preference between these?

BillMills:
Oh good point @s-good, we should use the gsw implementation and not add dependencies, unless there's a really good reason to add them - let me know what you and @BoyerWOD decide.

BoyerWOD:
TEOS-10 is probably best choice since it supersedes EOS-80.  I dont expect differences between the two would be noticeable in a stability check

s-good (approved):
Thanks - this looks great to me!

BillMills:
terrific - I want to get the CI test suite migrated to GH actions and running again before we merge here; I'll have time for that on Sunday and we can merge.

BillMills:
After fussing with migrating from travis to ghactions to get out unit testing back online, I noticed we actually have a helper for converting depth to pressure we use everywhere else. It's the exact same TEOS-10 via gsw as before, but changed to use it just for internal consistency. I'm going to go ahead and merge here, as always please let me know any further problems or requests.

